PR #11, opened 2015-04-05
Exactmass mod

divinebird:
Added method parameters to search MZpeaks including the mass defect.


tomas-pluskal:
Hi Vasily (I assume it's you),

Thank you for the contribution!

Could you please add your full name and public email address to your Github profile? We want to be able to contact all contributors in the future, in case of licensing changes or other issues.

Before merging your pull request, I want to ask two things:

- What is the reason for having two parameters (massDefect and absError)? Wouldn't a single parameter be sufficient? Generally, less parameters is better and more easily understandable for users.

- Since this function may only be used by some users, how about making this parameter optional? I recently added a new type OptionalParameter that could be used if you decide to merge massDefect and absError into a single parameter. If two parameters are needed, we could use OptionalModuleParameter. What do you think?

tomas-pluskal:
You can see here how the OptionalParameter is used:

https://github.com/mzmine/mzmine2/blob/master/src/main/java/net/sf/mzmine/modules/peaklistmethods/filtering/rowsfilter/RowsFilterParameters.java

divinebird:
Hi Tomáš,

Ok! I will see usage and will remake it.
What about reason. I done this modification for my friend. He looking for
some chemical material. I don't know what type of spectrometer he uses but
he get data. He needs to remove noise. Also he needs collect masses with
necessary mass defect. So if we bring some molecule some mass goes to bond
energy and we have the not integer mass. So as I understood friend like to
find some molecule with theoretical mass defect.
The tolerance is necessary because we need looking for interval. So exactly
we collect masses equals massDefect +/- absError)

Regards,
Vasily

чт, 9 апр. 2015 г. в 11:35, Tomáš Pluskal <notifications@github.com>:

You can see here how the OptionalParameter is used:

tomas-pluskal:
Hi Vasily,

Could you give some example of the kind of masses/molecules that you are detecting?
I have a feeling that your method might be more suitable as an identification algorithm (after peak detection), rather than as a filter in the exact mass detector. But please give me some example, then I may understand better.

Tomas

rspopov:
Hi Tomáš,
Vasily added this parameter at my request. The reason was as follows. We were studying steroid glycosides by LCMS. Resulted chromatograms were contaminated by peaks with mass defect 0.9 (possibly due to contamination by the products of decomposition PTFE at some stage), whereas studied glycosides have mass defect from 0.2 to 0.6. In addition, chromatogram builder found peaks with masses 1033.9 and 112.9 of calibrant solution. Generally after chromatogram builder stage such contaminant-peaks is one third of all peaks in peak list (picture). It is made it difficult to further processing and analyzing of data. Mass defect filter was added to remove such uninformative peaks. 

I understand that such case is some special, but I think that this filter may be used other users in their target metabolomic researches. For example, in studies of polar low-mass metabolites such as amino acids, mono- and disaccharides results may be refined by m/z window filtering as well as by mass defect filtering (mass defect of such metabolites not exceed 0.15). The filter will help remove of contamination and uninformative peaks in many other cases. 
peaks

tomas-pluskal:
Hi!
Thank you for the clarification.
As far as I understand, the peaks that you are referring to are real, existing peaks (not artificial noise in the data). 
The peaks represent ions that you are not interested in (in your experiment), nevertheless those are true, existing ions.
So, I think it is not the correct approach to filter these peaks during peak detection. Instead, I think would be more logical to detect all peaks and then process the resulting peak list. 
For example, check the "Peak list rows filter". We could add another field to this filter ("Mass defect") or create a similar filter specifically for this purpose.

What I still don't understand is why you need two parameters for this filter. Why not use only a single parameter of type RangeParameter, to define the range of allowed mass defects?

rspopov:
Yes, this peaks is associated to ions of compounds represented in sample.
Filter arrangement in the Mass detection module is based on the fact that the unnecessary data is deleted in the first stage and the program does not need processing uninformative data. But I think that your approach is more correct. Filtering by mass defect via a field in "Peak list rows filter" is more logical and clearly to users.
What about two parameters. Filter remove the peaks with mass that does not belong in the interval massDefect+/-absError. So we can not use one parameter mass defect. I'm not good at programming but as I understand in this case can be used type RangeParameter to define the interval (minmassDefect - maxmassDefect instead of massDefect+/-absError). I will contact with Vasily and we will try to fix this.

tomas-pluskal:
I'm closing this pull request for now..

PR #26, opened 2015-05-07
Added EMF and EPS save as options to TIC and XIC graphs

dyrlund:


dyrlund:
I set the target size of the export to depend on the component size on the screen to follow the implementation of the PNG export. Anyway, since the two new image formats are vector files, then the user can scale the images without loosing quality and I therefore think that we should keep the export simple without user defined width and height parameters.

I will change the implementation to account for the other issues you raised.

dyrlund:
With regards to the error message box for the user, will the following be enough or are you thinking of a message box with the error?
LOG.log(Level.WARNING, "Unable to save image.", e);

tomas-pluskal:
I thought about a message box using MZmineCore.getDesktop().displayErrorMessage(), because this is a purely GUI-based operation.

dyrlund:
Is there a nicer way to code this part of the code?

	       if (file.length() > 3) {
		   if (!file.substring(file.length() - 3).toLowerCase().equals("emf")) {
		       file = file + ".emf";
    		   }
	       }
	       else { file = file + ".emf"; }

tomas-pluskal:
how about
if (file.toLowerCase().endsWith(".emf")) file += ".emf";

dyrlund:
Yep - exactly what I was looking for. Very nice way to solve it :+1:

dyrlund:
I moved the saving of the images to a new class to run it in a separate thread.

tomas-pluskal:
I think the SaveImage class was not included in the commit..

dyrlund:
Wierd. I have added it now.

tomas-pluskal:
Looks good, merging now.

PR #36, opened 2015-05-28
Rserve integration 3.1

golgauth:
Migration of the baseline correction module to use Rserve.
Removed the conflicting file "HeatMapTask.java" we had both changed...


tomas-pluskal:
Looks good, I'll merge it. 
For future commits, coudl you please set your editor/IDE to indent using spaces instead of tabs?

golgauth:
Well, since I know exactly what files I modified, this would be easier to fix this at the and of my pull requests stack. Otherwise, I'll have more conflicts to deal with, I suppose (each branch expands from the previous one).

I gonna see if I get into trouble doing this now, or not.

golgauth:
By the way, I noticed you systematically quite unindent the code and wrap/limit the width of the text (what is the number of characters required?). It seems you're using eclipse as well: so may be you could give me the whole settings required for formatting before commit.

Thanks.

tomas-pluskal:
Please find attached the formatting style for Eclipse.

If you're using Eclipse, please format the sources before committing, and
also organize the imports (it's Command+Shift+O on Mac, not sure about
other platforms). It would be also nice if you cleanup the unused
(commented-out) code unless you are planning to use it in the future.

Tomas

golgauth:
Thanks, I'll apply those settings to eclipse (I already did some changes for tabs to spaces, ...).
Regarding the commented lines, I usually cleanup as much as I can, except for things that I assume not completely done (or could be better). So, in theory, this will be completely clean a bit later ;)

golgauth:
Note about this pull request (3.1): The code shouldn't compile due to HeatMapTask.java being outdated (struggling with Git, sorry). You'll find the right one in next pull request (4.1) or in my "rserve-integration-4.1" branch.

PR #37, opened 2015-05-29
Rserve integration 4.1

golgauth:
Moved peaklists deconvolution module to use Rserve.


src/main/java/net/sf/mzmine/modules/peaklistmethods/peakpicking/deconvolution/DeconvolutionTask.java
```diff
@@ -59,6 +61,11 @@
     // User parameters
     private final ParameterSet parameters;
 
+	private RSessionWrapper rSession;
+	private String errorMsg;
+	private boolean userCanceled;
```
tomas-pluskal: Instead of this new userCanceled variable you can simply use the isCanceled() method

src/main/java/net/sf/mzmine/modules/peaklistmethods/peakpicking/deconvolution/DeconvolutionTask.java
```diff
@@ -59,6 +61,11 @@
     // User parameters
     private final ParameterSet parameters;
 
+	private RSessionWrapper rSession;
```
tomas-pluskal: We have several deconvolution methods, but so far only one of them (xcms-based) depends on R.
So I think it makes more sense to move R-related variables into the xcms deconvolution class, instead of this (general) DeconvolutionTask. If we keep it here, there is a slight risk that a problem in R will bring down the whole deconvolution module.

src/main/java/net/sf/mzmine/modules/peaklistmethods/peakpicking/deconvolution/DeconvolutionTask.java
golgauth: Maybe. This is not the same meaning. I think this variable is used in BaselineDeconvPreviewSetup, in case the user presses escape, or else. Anyway, in the context of a preview is available for the given feature, this variable is necessary to tell: "the user asked for cancelling the task" without triggering the Event (=calling @override cancel()). I should probably look deeper into this, because it might be something that was necessary at a time but is not anymore due to Model changes, or else...

But, "userCanceled" and "isCanceled()" are two different things.

src/main/java/net/sf/mzmine/modules/peaklistmethods/peakpicking/deconvolution/DeconvolutionTask.java
golgauth: Well, I know that this isn't marvelous from the design point of view. But, R isn't even called in the features that do not require it. At first, I did as you you say, calling R stuffs only from inside the CentWave.resolvePeak(), but the thing is that resolvePeak() is called many times, and forking Rserve several times is not consistent. It was at the time the JRI engine reference was a singleton instantiated once and for all. Now, the Rsession.open() has to reside outside of the resolvePeak()...

I could have a look again, but, I am not sure about how to do better here.

src/main/java/net/sf/mzmine/modules/peaklistmethods/peakpicking/deconvolution/DeconvolutionTask.java
golgauth: Seems you are right. Without this "race condition" constraint due to the Swing progress bar thread (Cf. BaselineCorrectorSetupDialog), that is to say, in every other cases, this additional variable seems useless. Let's remove it...

golgauth: Done!

src/main/java/net/sf/mzmine/modules/peaklistmethods/peakpicking/deconvolution/DeconvolutionTask.java
tomas-pluskal: If I am not mistaken, CentWave is a singleton itself, instantiated in DeconvolutionParameters:
```
    private static final PeakResolver[] RESOLVERS = {
	    new BaselinePeakDetector(), new NoiseAmplitudePeakDetector(),
	    new SavitzkyGolayPeakDetector(), new MinimumSearchPeakDetector(),
	    new CentWaveDetector() };
```

So although you are right that resolvePeaks() is called many times, the R-related properties could be initialized in the CentWaveDetector's constructor... or not?

golgauth: Well, at very first sight, no. The session (or Rserve fork process) I start via (RSessionWrapper.open()), is open on demand, then closed when the task is done. Since we are using some kind of TCP/IP protocol. I do not want to compromise the entire application by instantiating a CentWave dedicated Rserve session once and for all as we used to do with JRI. That is not the policy ;)... The strategy is to keep the app working even if something bad happen during a computation (Rserve session crashed, eval failed, connection failed, parallel tasks are taking all CPU available...). If the CentWave deconvolution that just ran crashed/blocked for whatever reason, I want to be burn proof and able to run another deconvolution right after the failure, having MZmine still behaving well in the meanwhile.

I'll have a closer new fresh look at MZmine's model (it's been a while I haven't), but I am pretty sure we can't do anything even tricky one from the resolver's constructor...

golgauth: Would that sort of "Template method pattern" oriented approach please 
you more?

https://github.com/golgauth/mzmine2/tree/rserve-integration-4/src/main/java/net/sf/mzmine/modules/peaklistmethods/peakpicking/deconvolution

This is more or less elegant way of keeping all R-related stuffs 
self-contained in the CentWaveDetector.

(Not yet pushed on the rserve-integration-**4.1** branch)

Readable diff in attachment.


On 05/29/2015 03:05 PM, Tomáš Pluskal wrote:
>
> In 
> src/main/java/net/sf/mzmine/modules/peaklistmethods/peakpicking/deconvolution/DeconvolutionTask.java 
> <https://github.com/mzmine/mzmine2/pull/37#discussion_r31322097>:
>
> > @@ -59,6 +61,11 @@
> >      // User parameters
> >      private final ParameterSet parameters;
> >
> > +	private RSessionWrapper rSession;
>
> If I am not mistaken, CentWave is a singleton itself, instantiated in 
> DeconvolutionParameters:
>
> |     private static final PeakResolver[] RESOLVERS = {
>          new BaselinePeakDetector(), new NoiseAmplitudePeakDetector(),
>          new SavitzkyGolayPeakDetector(), new MinimumSearchPeakDetector(),
>          new CentWaveDetector() };
> |
>
> So although you are right that resolvePeaks() is called many times, 
> the R-related properties could be initialized in the 
> CentWaveDetector's constructor... or not?
>
> —
> Reply to this email directly or view it on GitHub 
> <https://github.com/mzmine/mzmine2/pull/37/files#r31322097>.
>


-- 
Gauthier BOAGLIO
CEFE - UMR 5175
1919 route de Mende
F-34293 Montpellier cedex 5

Tel: +33/0 4 67 61 32 15
Fax: +33/0 4 67 61 33 36

email: gauthier.boaglio@cefe.cnrs.fr
www:   http://www.cefe.cnrs.fr/en/evolutionary-ecology-and-epidemiology/gauthier-boaglio
        http://www.evolepid.org/people.php?name=boaglio

tomas-pluskal: To be honest, I liked the previous solution better. Adding more methods to the PeakResolver interface just complicates things.. and those Object... args look rather confusing. 

I will merge the branch as it is now :)

golgauth: Ok, up to you ;). Your second argments is understandable (a get that those two hooks "before/after batch resolution" with undetermined arguments are huge bags were anything can be put into, making their purpose unclear - too much freedom for serving a purpose that is useful to only one among the Resolvers). The first one, a little bit less since we add 3 (R-dedicated) methods in the PeakResolver as well. Moreover those methods added to a generic interface that is not particularily designed to use R in a systematic way...

But that's ok for me :)

PR #38, opened 2015-05-29
Rserve integration 5

golgauth:
Moved CAMERA search module to use Rserve.


tomas-pluskal:
Did you apply the Eclipse formatting style I sent you? 
It seems all the files in this pull request are formatted with tabs instead of spaces.

golgauth:
I did, but not the file changes related to this particular pull request, which was anterior to that matter.

PR #43, opened 2015-06-01
Rserve integration 7.1

golgauth:
Removed all JRI references + formatting.

### DIFF
--- a/src/main/scripts/startMZmine_Windows.bat
+++ b/src/main/scripts/startMZmine_Windows.bat
@@ -52,16 +52,6 @@ mkdir %TMP_FILE_DIRECTORY%
 
 rem Set R environment variables.
 set R_HOME=C:\Program Files\R\R-2.12.0
-set R_SHARE_DIR=%R_HOME%\share 
-set R_INCLUDE_DIR=%R_HOME%\include
-set R_DOC_DIR=%R_HOME%\doc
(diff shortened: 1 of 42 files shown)


src/main/scripts/startMZmine_Windows.bat
```diff
@@ -52,16 +52,6 @@ mkdir %TMP_FILE_DIRECTORY%
 
 rem Set R environment variables.
 set R_HOME=C:\Program Files\R\R-2.12.0
```
tomas-pluskal: Is the R_HOME variable still required? 
I thought you had implemented some auto-detection of R location, haven't you?

golgauth: Yes, I did. But in some cases it can be required (if auto-detection fails). That is why I kept it for now. At the time I started working on all this you had an excellent suggestion which was more or less: "Remove it completely from the scripts, and, in case auto-detection failed, ask the user to give a path via the Project > Set Preferences panel".

I had no time to implement it, but the idea still stands (got it written on one of my several post-it :).

golgauth:
Done with the R_HOME env var removal.

golgauth:
Made some cleanup and very slight arrangements. That should be all for now regarding this pull request. Unless you have more comments...

PR #44, opened 2015-06-01
Tracking of GUI and batch usage

dyrlund:


src/main/scripts/startMZmine_Linux.sh
```diff
@@ -29,6 +29,11 @@ TMP_FILE_DIRECTORY=/tmp
 export TMP_FILE_DIRECTORY=$TMP_FILE_DIRECTORY/$MZMINE_UNID
 mkdir $TMP_FILE_DIRECTORY
 
+# If you do not want MZmine to sends anonymous statistics of the module
+# usage in the software then please set the MZ_STATISTICS parameter to 0.
+# 1 = activated, 0 = deactivated.
+export MZ_STATISTICS=1
```
tomas-pluskal: A minor point: can we name it MZMINE_STATISTICS instead of MZ_STATISTICS?

src/main/java/net/sf/mzmine/main/MZmineCore.java
```diff
@@ -220,6 +220,10 @@ public void run() {
 	    NewVersionCheck NVC = new NewVersionCheck(CheckType.DESKTOP);
 	    new Thread(NVC).start();
 
+	    // Tracker
+	    GoogleAnalyticsTracker GAT = new GoogleAnalyticsTracker("GUI Loaded", "/JAVA/Main/GUI");
+	    new Thread(GAT).start();
```
tomas-pluskal: The Thread class has a function setPriority().
I think it could be a good idea to set the priority of these two threads (GAT and NVC) to the lowest possible value.

tomas-pluskal:
Hi Thomas, 
I think there is a (minor) conflict with Gauthier's changes (removing R_HOME from the startMZmine scripts), so I cannot merge the pull request now. Please take a look at it.

dyrlund:
Moved to pull request #47

PR #48, opened 2015-06-02
Removed JRI deps from POM.

golgauth:


tomas-pluskal:
Thanks!
I guess we can then also completely remove the src/lib/rJava folder?

golgauth:
Yep. Good thinking. Done!

tomas-pluskal:
Thanks! Awesome work!

golgauth:
Well, I hope so. We'll have to test it intensively...
I tried another approach for the "Deconvolution module design" but this is not that interesting, in the end. In case you're curious: https://github.com/golgauth/mzmine2/tree/rserve-integration-6.1/src/main/java/net/sf/mzmine/modules/peaklistmethods/peakpicking/deconvolution
Cheers.

dyrlund:
I did some tests of the new Rserve and it works great so far. Really nice work Gauthier! I expect that this will reduce the issues many users experience with the R integration. And it is so nice that the R dependent modules can now be multi-threaded.

golgauth:
Glad to here that! I had it in the tracks for while and was looking forward to find time to pull it into the main trunk (which was moving faster than me). And the jet lag with Japan (I mean Tomas) is pretty much "not in our favor", unless you wake up early ;).

Hope not too many bug reports will blow up...

PR #49, opened 2015-06-02
Added tracking of tasks to the TaskController

dyrlund:


src/main/java/net/sf/mzmine/taskcontrol/impl/TaskControllerImpl.java
```diff
@@ -92,6 +93,14 @@ public void addTasks(Task tasks[]) {
     }
 
     public void addTasks(Task tasks[], TaskPriority priority) {
+	// Track module usage
+	String taskClass = tasks[0].getClass().getName();
```
tomas-pluskal: tasks[0] is going to throw an exception if the array is empty (see the code comment a few lines below). 
Generally in Java, using anything[0] signifies there is a programming problem.
In this case, you should probably go through the array and make a Set of all Classes of the tasks - because there is no guarantee that the tasks[] array contains tasks of the same class.

dyrlund: Right, I wasn't aware that tasks[0] could be empty.

src/main/java/net/sf/mzmine/taskcontrol/impl/TaskControllerImpl.java
```diff
@@ -92,17 +96,31 @@ public void addTasks(Task tasks[]) {
...
-
 	// It can sometimes happen during a batch that no tasks are actually
 	// executed --> tasks[] array may be empty
 	if ((tasks == null) || (tasks.length == 0))
 	    return;
 
+	List<String> taskClass = new ArrayList<String>();
+	String taskClassName;
 	for (Task task : tasks) {
+	    taskClassName = tasks[0].getClass().getName();
```
tomas-pluskal: It still has tasks[0]  :)

dyrlund: Ups :-) Well spotted :+1:

PR #50, opened 2015-06-03
Cleanup of startMZmine_Windows.bat

dyrlund:
Please comment on the changes I have made to the Windows script. Once we are happy with it, I will make the same changes to the linux ans mac scripts and you can then also comment on them.

### DIFF
--- a/src/main/scripts/startMZmine_Linux.sh
+++ b/src/main/scripts/startMZmine_Linux.sh
@@ -39,7 +39,7 @@ echo "Checking physical memory size..."
 TOTAL_MEMORY=`free -b | awk '/Mem:/ { print int($2 / 1024^2) }'`
 echo "Found $TOTAL_MEMORY MB memory"
 
-if [ "$HEAP_SIZE" -eq AUTO ]; then
+if [ "$HEAP_SIZE" == "AUTO" ]; then
   if [ "$TOTAL_MEMORY" -gt 4096 ]; then
 	HEAP_SIZE=`expr $TOTAL_MEMORY - 2048`
   else
--- a/src/main/scripts/startMZmine_Linux.sh
+++ b/src/main/scripts/startMZmine_Linux.sh
@@ -39,7 +39,7 @@ echo "Checking physical memory size..."
 TOTAL_MEMORY=`free -b | awk '/Mem:/ { print int($2 / 1024^2) }'`
 echo "Found $TOTAL_MEMORY MB memory"
 
-if [ "$HEAP_SIZE" == "AUTO" ]; then
+if [ "$HEAP_SIZE" = "AUTO" ]; then
   if [ "$TOTAL_MEMORY" -gt 4096 ]; then
 	HEAP_SIZE=`expr $TOTAL_MEMORY - 2048`
   else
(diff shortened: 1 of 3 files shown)


src/main/scripts/startMZmine_Windows.bat
```diff
@@ -1,12 +1,44 @@
 @echo off
 
-rem This is necessary to access the TOTAL_MEMORY and ADDRESS_WIDTH variables inside the IF block
+:: *****************************************
+:: Optional values - Please modify as needed
+:: *****************************************
+
+:: Total amount of memory in MB available to MZmine 2.
+:: 0 = automatically determined
```
tomas-pluskal: the comment says 0 = automatically determined, but below AUTO is used instead of 0.

dyrlund:
I am not 100% sure about this line:
if [ "$HEAP_SIZE" -eq AUTO ]; then

tomas-pluskal:
The -eq AUTO is not going to work, because -eq is used to compare integers:

     n1 -eq n2     True if the integers n1 and n2 are algebraically equal.

The = operator might be better:

    s1 = s2       True if the strings s1 and s2 are identical.

But you'll need to add parenthesis around AUTO.

src/main/scripts/startMZmine_Windows.bat
```diff
@@ -1,12 +1,44 @@
...
+set TMP_FILE_DIRECTORY=%TEMP%
+
+:: Allow MZmine to send anonymous statistics on the module usage? 
+:: 1 = activated, 0 = deactivated
+:: Default: 1
+set MZMINE_STATISTICS=1
+
+:: It is usually not necessary to modify the JAVA_COMMAND parameter, but if you like to run
+:: a specific Java Virtual Machine, you may set the path to the java command of that JVM
+set JAVA_COMMAND=java
```
tomas-pluskal: Since you're cleaning up the scripts, maybe we could change this to 
JAVA_COMMAND=java.exe
in the Windows script. That will make the meaning more obvious. The comment can be also updated ("..., you may set the path to the java.exe command...").

dyrlund:
I changes the comparison to:
f [ "$HEAP_SIZE" == "AUTO" ]; then

src/main/scripts/startMZmine_Linux.sh
```diff
@@ -1,48 +1,70 @@
...
+# ***********************************
+
+# By default we set the HEAP_SIZE to 1024 MB on 32-bit systems. On 64-bit systems we 
+# either set it to half of the total memory or 2048 MB less than the total memory.
 echo "Checking physical memory size..."
 TOTAL_MEMORY=`free -b | awk '/Mem:/ { print int($2 / 1024^2) }'`
 echo "Found $TOTAL_MEMORY MB memory"
 
-if [ "$TOTAL_MEMORY" -gt 4096 ]; then
+if [ "$HEAP_SIZE" == "AUTO" ]; then
```
tomas-pluskal: Hi Thomas,
I think it should be a single equal sign (=), not ==.
As I posted in the previous message:

     s1 = s2       True if the strings s1 and s2 are identical.

PR #55, opened 2015-07-02
Fixed issue with newly opened peak list tables being hidden behind main window

dyrlund:
Thanks to Tomas for solving this issue :+1:


tomas-pluskal:
Did you find any other occurrences of mousePressed() in the source codes?

dyrlund:
Yes, but none which I think should be changed since they handled drag events.

PR #60, opened 2015-07-09
Removed the "hidden m/z range" hack

tomas-pluskal:
Thomas,
I changed the Range component of the parameter setup dialog so it automatically rounds the min value down and the max value up. With this change, I believe your "hidden m/z range" hack is not needed anymore, because the original m/z range will never be restricted by rounding.
Take a look.


src/main/java/net/sf/mzmine/parameters/parametertypes/ranges/DoubleRangeComponent.java
```diff
@@ -73,8 +74,12 @@ public void setNumberFormat(NumberFormat format) {
...
     public void setValue(Range<Double> value) {
-        minTxtField.setText(format.format(value.lowerEndpoint()));
-        maxTxtField.setText(format.format(value.upperEndpoint()));
+        NumberFormat floorFormat = (NumberFormat) format.clone();
+        floorFormat.setRoundingMode(RoundingMode.FLOOR);
+        NumberFormat ceilFormat = (NumberFormat) format.clone();
+        ceilFormat.setRoundingMode(RoundingMode.CEILING);
+        minTxtField.setText(floorFormat.format(value.lowerEndpoint()));
+        maxTxtField.setText(ceilFormat.format(value.upperEndpoint()));
     }
```
dyrlund: Won't this affect the value of all range parameters set in any dialog window - e.g. also if a user specifies their own range?

tomas-pluskal: No - the rounding was always there, I just adjusted the direction of the rounding.

PR #61, opened 2015-07-10
Fixed issue with peak information being shown twice in the legend

dyrlund:


src/main/java/net/sf/mzmine/modules/visualization/tic/TICPlot.java
```diff
@@ -681,7 +681,7 @@ public String generateLabel(final XYDataset xyDataSet,
 		}
 	    });
 
-	    numOfPeaks++;
+	   // numOfPeaks++;
```
tomas-pluskal: I am not a big fan of such commented-out one-liners. Soon nobody is going to remember why this line is commented out. If the code is not needed, please remove it completely. If there is a reason to keep it, the reason should be explained in a comment.

dyrlund: I agree - I was just not sure if there was a reason for increasing the variable at the specific position.

PR #62, opened 2015-07-20
IDA Visualizer

dyrlund:
Visualizer which plots the data from an Information Dependent Acquisition (IDA) experiment. The x-axis corresponds to retention time of the ms/ms spectrum and the y-axis is the m/z value of the precursor ion.


src/main/java/net/sf/mzmine/modules/visualization/ida/IDADataSet.java
```diff
@@ -0,0 +1,377 @@
...
+
+    public Number getZ(int series, int item) {
+	return intensityValues[item];
+    }
+
+    public Number getMaxZ() {
+	return maxIntensity;
+    }
+
+    public Number getMaxZ(Range<Double> mzRange) {
```
dyrlund: Is there a way to optimize this function?

src/main/java/net/sf/mzmine/modules/visualization/ida/IDADataSet.java
```diff
@@ -0,0 +1,377 @@
...
+	for (int row = 0; row < totalmsmsScans; row++) {
+	    if (mzRange.contains(mzValues[row])) {
+		if (max < intensityValues[row]) {
+		    max = intensityValues[row];
+		}
+	    }
+	}
+	return max;
+    }
+
```
dyrlund: Is there a way to optimize this function?

PR #63, opened 2015-07-21
Fixed null point error in the standard compound normalizer

dyrlund:


src/main/java/net/sf/mzmine/modules/peaklistmethods/normalization/standardcompound/StandardCompoundNormalizerTask.java
```diff
@@ -190,10 +195,7 @@ public void run() {
 
 			Feature standardPeak = standardRow.getPeak(file);
 			if (standardPeak == null) {
-			    // What to do if standard peak is not
-			    // available? (Currently this is ruled out by the
-			    // setup dialog, which shows only peaks that are
-			    // present in all samples)
+			    // What to do if standard peak is not available?
```
dyrlund: The setup dialog does not only show peaks which are present in all samples, which is why the null check is needed. But should the dialog only show the peaks which are present in all samples?

dyrlund: I just checked the help file for the module and it says that:
"... User can select one or multiple internal standard peaks, which must be present in all raw data files."

So I guess that there is something not working with the setup dialog.

PR #68, opened 2015-08-03
Added search for neutral loss ions to the IDA visualizer

dyrlund:


src/main/java/net/sf/mzmine/modules/visualization/ida/IDADataSet.java
```diff
@@ -273,7 +273,7 @@ public void setColor(int series, int item, Color c) {
      * 
      */
     public void highlightSpectra(double mz, double ppm, int minIntensity,
```
tomas-pluskal: This comment is not related to this commit, but I wonder why you used double mz and double ppm in the constructor. Other modules use the MZTolerance class specifically for this purpose.

dyrlund: Because I am only using ppm tolerance and not m/z tolerance.

tomas-pluskal: Let me rephrase the question: why do you prefer to use only ppm tolerance in this case, when all other MZmine modules are using the MZTolerance class that can specify both absolute and relative tolerances?

dyrlund:
I implemented it in that way since I didn't know there was a MZTolerance class :) I'll update the code so that it will use the MZTolerance class instead ...

dyrlund:
I have now changed to the MZTolerance class instead.

I do however not like that I have to define both the m/z tolerance and the ppm tolerance for this parameter. Is there a reason why this is the way it is implemented? I think it would make more sense if the user could e.g. decide to only input the ppm tolerance and then leave the m/z tolerance empty.

tomas-pluskal:
You don't need to define both - the tolerance is always the maximum of the two values. If you set ppm tolerance to 0, only m/z tolerance will be used, and vice versa.

There are practical reasons for this implementation. Instrument accuracy is typically defined at certain m/z - e.g., 5 ppm at 400 m/z. But when you to lower m/z values (e.g. 100 m/z) the 5 ppm becomes unrealistic. That's where the absolute m/z tolerance comes in - you can always define minimal tolerance window.

PR #71, opened 2015-08-04
Changes to the IDA Visualizer

dyrlund:
- Values were set for intensityType, normalizationType and minPeakInt
- ppm tolerance was changed to the MZTolerance parameter


src/main/java/net/sf/mzmine/modules/visualization/ida/IDADataSet.java
```diff
@@ -274,12 +275,11 @@ public void setColor(int series, int item, Color c) {
      *            color.
      * 
      */
-    public void highlightSpectra(double mz, double ppm, int minIntensity,
```
tomas-pluskal: By the way, we normally use double type for intensity, I see minIntensity is int here. Is there any reason for that? Remember that int has a fixed range of allowed values -2,147,483,648 to 2,147,483,647, so if you get to intensities above 2E9 you might get overflows (2E9 values are commonly seen e.g. on Q Exactive).

dyrlund: I have changed it to a double instead.

PR #75, opened 2015-08-06
Update to progress tracking for loading a project

dyrlund:
When loading a project, is there a reason for why:
1. Raw data files and peak lists are not loaded into the lists as they are loaded from the zip file?
2. The progress tracker does not keep track of the entire project load?

I have made some changes to account for the above, please add your comments.

The issue is the same for saving a project where the total progress is not tracked. For me this is not very convenient.

### DIFF
--- a/src/main/java/net/sf/mzmine/modules/projectmethods/projectload/MyZIS.java
+++ b/src/main/java/net/sf/mzmine/modules/projectmethods/projectload/MyZIS.java
@@ -1,0 +1,40 @@
+/*
+ * Copyright 2006-2015 The MZmine 2 Development Team
+ *
+ * This file is part of MZmine 2.
+ *
+ * MZmine 2 is free software; you can redistribute it and/or modify it under the
+ * terms of the GNU General Public License as published by the Free Software
+ * Foundation; either version 2 of the License, or (at your option) any later
+ * version.
+ *
+ * MZmine 2 is distributed in the hope that it will be useful, but WITHOUT ANY
+ * WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS FOR
+ * A PARTICULAR PURPOSE. See the GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License along with
+ * MZmine 2; if not, write to the Free Software Foundation, Inc., 51 Franklin
+ * St, Fifth Floor, Boston, MA 02110-1301 USA
+ */
+
+package net.sf.mzmine.modules.projectmethods.projectload;
+
+import java.io.IOException;
+import java.io.InputStream;
+import java.util.zip.ZipInputStream;
+
+public class MyZIS extends ZipInputStream {
+
+    public MyZIS(InputStream in) {
+        super(in);
+    }
+
+    @Override
+    public void close() throws IOException {
+    }
+
+    public void closeZIS() throws IOException {
+        super.close();
+    }
+
+}
--- a/src/main/java/net/sf/mzmine/modules/projectmethods/projectload/ProjectOpeningTask.java
+++ b/src/main/java/net/sf/mzmine/modules/projectmethods/projectload/ProjectOpeningTask.java
@@ -26,7 +26,6 @@ import java.io.FileOutputStream;
 import java.io.IOException;
 import java.io.InputStream;
 import java.io.InputStreamReader;
-import java.util.Enumeration;
 import java.util.Hashtable;
 import java.util.logging.Logger;
 import java.util.regex.Matcher;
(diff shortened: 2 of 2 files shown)


tomas-pluskal:
1. Raw data files and peak lists are not loaded into the lists as they are loaded from the zip file?

The reason for this is to have project loading as an atomic operation that can also be canceled. Imagine you already have some data in the current project. Then you start loading a different project, and cancel the loading task in the middle. You will end up with a mess. That's why we load the project in a separate space and only replace the project once the loading is complete.

2. The progress tracker does not keep track of the entire project load? 

The current implementation was written by Sandra and I agree it's a bit confusing. Instead of defining an arbitrary 50:50 ratio, though, I think it would be better to use a ZipInputStream instead of the ZipFile that is used currently. You can use e.g. the [CountingInputStream](http://docs.guava-libraries.googlecode.com/git/javadoc/com/google/common/io/CountingInputStream.html) from the Guava library to track how much was already read. I mean something like this:

```
FileInputStream fis = new FileInputStream(file);
CountingInputStream cis = new CountingInputStream(fis);
ZipInputStream zis = new ZipInputStream (cis);
```

dyrlund:
I see your point with regards to cancelling the loading of a project. But how often does this happen? I think that sometimes it is also a question of making things as user friendly as possible and then sacrifice some features. But if you feel that we should leave the loading as one operation then we will do so.

I'll work on changing the progress tracker to use the ZipInputStream instead.

tomas-pluskal:
You are probably right. Maybe we could simply display a warning in case the current project contains some data, and then immediately replace the current project with a new project instance and start loading the project contents into the new instance. That would work.

dyrlund:
I have added a warning before replacing the data. I like that implementation. What do you think?

tomas-pluskal:
OK, good. Are you still planning to add the CountingInputStream for the progress tracking?

dyrlund:
Yes :)

src/main/java/net/sf/mzmine/modules/projectmethods/projectload/ProjectOpeningTask.java
```diff
@@ -121,68 +128,108 @@ public double getFinishedPercentage() {
...
-	    ProjectManager projectManager = MZmineCore.getProjectManager();
-	    projectManager.setCurrentProject(newProject);
-
-	    logger.info("Finished opening project " + openFile);
-
-	    setStatus(TaskStatus.FINISHED);
+            // Check if existing raw data files are present
+            ProjectManager projectManager = MZmineCore.getProjectManager();
+            if (projectManager.getCurrentProject().getDataFiles().length > 0) {
+                System.out.println(projectManager.getCurrentProject()
```
tomas-pluskal: The println() call could be removed...

dyrlund:
How would you update the ZipInputStream based on the data loaded from the zip file? In the current version, everything is processed based on the following:
ZipFile zipFile = new ZipFile(openFile);

From the ZipInputStream I can only seem to get the next entry - getNextEntry() - and I therefore don't have the advanced functions present which I have in the ZipFile. Any suggestions?

tomas-pluskal:
The getNextEntry() method is sufficient to get all the data.
But we will have to change the signatures of the load....() methods

dyrlund:
I added the initial work on replacing zipFile with ZipInputStream - I won't get more done today. I have two issues which still have to be solved:

1.
Only the first peak list is loaded, after that the zis stream is closed. If I exclude the below line from the code then the error does not occur. So the problems seems to be related to the readPeakList closing the stream. Any idea why?
PeakList newPeakList = peakListOpenHandler.readPeakList(peakListStream);

2.
The ZipFile is still used to load the raw data files.

tomas-pluskal:
Apparently the SaxParser automatically closes the stream. You can avoid that, see e.g. the second answer here: http://stackoverflow.com/questions/8341680/java-saxparser-keep-inputstream-open

dyrlund:
Yep, I saw that also. I already fixed that :)

src/main/java/net/sf/mzmine/modules/projectmethods/projectload/MyZIS.java
```diff
@@ -0,0 +1,40 @@
...
+ * St, Fifth Floor, Boston, MA 02110-1301 USA
+ */
+
+package net.sf.mzmine.modules.projectmethods.projectload;
+
+import java.io.IOException;
+import java.io.InputStream;
+import java.util.zip.ZipInputStream;
+
+public class MyZIS extends ZipInputStream {
```
tomas-pluskal: I'd recommend a more descriptive name for this class - e.g. UncloseableZipStream or something on that line.. remember somebody might be reading your code after you and a line like

MyZIS zis = new MyZIS(); 

is not so easy to understand, so people have to dig into the class sources.

dyrlund: Sounds fine with me to update the name of the class.

tomas-pluskal:
Replaced by pull request #81

PR #76, opened 2015-08-06
Disabled tracking for developers version

dyrlund:


src/main/java/net/sf/mzmine/main/GoogleAnalyticsTracker.java
```diff
@@ -50,9 +50,15 @@ public GoogleAnalyticsTracker(String title, String url) {
     }
 
     public void run() {
-	
+
 	// Only send data if SendStatistics variable is not set to 0
 	Boolean SendStatistics = MZmineCore.getConfiguration().getSendStatistics();
```
tomas-pluskal: As a convention, variable names in Java start with a lower-case letter. Can you please change SendStatistics -> sendStatistics?

PR #81, opened 2015-08-13
Project loading tracking

tomas-pluskal:
Thomas,
I reinstated the ZipFile, but added the CountingInputStream to track the progress using number of uncompressed bytes. This finally seems to work nicely.


dyrlund:
Perfect, great work! I think the implementation is a lot better than the previous :) We should also have a look at the project save module where the tracking is also wrong.

dyrlund:
Perfect, great work! I think the implementation is a lot better than the previous :) We should also have a look at the project save module where the tracking is also wrong.

PR #82, opened 2015-08-19
Added FWHM, Tailing factor and Asymmetry factor to peak lists

dyrlund:


src/main/java/net/sf/mzmine/modules/masslistmethods/chromatogrambuilder/ChromatogramBuilderTask.java
```diff
@@ -171,6 +172,9 @@ public void run() {
 	// Add new peaklist to the project
 	project.addPeakList(newPeakList);
 
+        // Add quality parameters to peaks
+        new QualityParameters(newPeakList);
```
tomas-pluskal: what is the meaning of this line? 
it looks a bit strange to just make an instance of a class and not assign the instance to any variable.

dyrlund: I moved the calculation of the quality parameters for a peak to a separate class which I can then call instead of having to add the code to all modules generating peak lists.

The reason that I don't assign the instance to any variable is that I don't need to do anything to the instance after the first run. Should I implement it in a different way?

tomas-pluskal: Yes, this way looks a bit strange as there is no need for the instance. This is normally done using a static method, e.g. 

public class QualityParameters {
    public static void calculateQualityParams(PeakList peakList) {
     ....
    }
}

and then

QualityParameters.calculateQualityParams(peakList);

dyrlund: As suggested, I have corrected so that a static method is used instead.

PR #139, opened 2015-12-12
Removed remaining rJava references from help files.

golgauth:
`rJava` references replaced by `Rserve` ones.

### DIFF
--- a/src/main/java/net/sf/mzmine/desktop/preferences/help/help.html
+++ b/src/main/java/net/sf/mzmine/desktop/preferences/help/help.html
@@ -22,10 +22,16 @@
 <dd>Format of intensity values.</dd>
 
 <dt>Number of concurrently running tasks</dt>
-<dd>Maximum number of tasks running simultaneously</dd>
+<dd>Maximum number of tasks running simultaneously.</dd>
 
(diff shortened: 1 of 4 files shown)


golgauth:
Hi T(h)omas(es) (@tomas-pluskal @dyrlund),

Just wondering: You remove completely the R_HOME environment variable from startup scripts (even in a commented way, since we have and automated detection).

The `R_HOME` environment variable is barely mentioned in the "Usage and workflow" section. Nevertheless, I suppose people who are using MZmine in _Headless mode_ on computing clusters, by ex., would need to have this information.

Don't you think we could be a bit more explicit about the usage, here?

tomas-pluskal:
Hi Gauthier, 
so far nobody reported any problems with the automatic R_HOME detection, so I don't think it is a big issue...

golgauth:
OK, I agree. It is fine with me too, if no one complaint so far ;)

Just as a reminder (mainly for myself) on how MZmine works for finding R executable:

1. Look for R location setting in "Project > Preferences".
2. Look for `R_HOME` environment variable (from startup script)
3. Autodetection: Attempt to find it from registry or well known binary locations (depending on the platform)

PR #150, opened 2015-12-17
Removed the peaks selection from PCA parameters - do we need it?

tomas-pluskal:
Hi Thomas @dyrlund , Sandra @SandraCastilloPriego,
I have removed the peak selection from PCA plot parameters. I think it is a bit confusing and normally users will always want to do PCA on the whole dataset. Please let me know if you have a different opinion.


dyrlund:
I'm happy with removing it.

PR #171, opened 2016-01-21
Extension of the MassBank search

tsufz:
The MassBank search does not consider both main MassBank servers (massbank.jp, massbank.ufz.de). The European MassBank is more environmental oriented and provides some different databases, for example mass spectra from environmental samples or literature. The Japanese MassBank related more to pure metabolomics spectra, but also includes the environmental standard spectra mirrored from European MassBank.

The provision of different searches for MassBank Japan / Europe is also recommended to avoid latencies. An alternative could be an optional selection of one of the both MassBanks.

Additionally, I updated the instrument type lists (and ordered them ascending).

### DIFF
--- a/src/main/java/net/sf/mzmine/modules/peaklistmethods/identification/onlinedbsearch/OnlineDatabase.java
+++ b/src/main/java/net/sf/mzmine/modules/peaklistmethods/identification/onlinedbsearch/OnlineDatabase.java
@@ -27,7 +27,8 @@ import net.sf.mzmine.modules.peaklistmethods.identification.onlinedbsearch.datab
 import net.sf.mzmine.modules.peaklistmethods.identification.onlinedbsearch.databases.HMDBGateway;
 import net.sf.mzmine.modules.peaklistmethods.identification.onlinedbsearch.databases.KEGGGateway;
 import net.sf.mzmine.modules.peaklistmethods.identification.onlinedbsearch.databases.LipidMapsGateway;
-import net.sf.mzmine.modules.peaklistmethods.identification.onlinedbsearch.databases.MassBankGateway;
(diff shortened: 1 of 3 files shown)


tomas-pluskal:
Hi Tobias,

Thanks a lot for the contribution! Are you interested in developing this module further? We currently don't have any spectral search function (we only search for the precursor mass). I you would be interested in implementing that, it would be most welcome.

Tomas

tsufz:
Hi Thomas,
I am not so familiar with Java programming. However, I could try to do something to get it implemented. An inter-linkage with MassBank of North America could be also interesting (http://mona.fiehnlab.ucdavis.edu/#/). Another option could be to use the splash key (http://splash.fiehnlab.ucdavis.edu/). We introduced it now to the MassBank records, but needs a complementary structure in mzMine. On the other hand, I would not loose too much time in developments regarding old MassBank software. We are in a more or less hold stage with further development, but we may switch to another platform at a curtain point.... (watch https://github.com/MassBank).

I consider to join Metabolomics 2016. We may organise a meeting. I dunno, if they plan a Hackathon as in 2015, but there was the big one in Dagstuhl...

tomas-pluskal:
Hi Tobias,

Indeed, I was at Dagstuhl, and I am also among the authors of the splash key, so it will be added to MZmine soon.

As for MoNA, they promised to provide a Java interface, after which it can be easily added to MZmine. Feel free to make additional pressure on them (e.g. by submitting in issue at https://bitbucket.org/fiehnlab/mona/issues/). 

I thought you might be interested in adding a MassBank spectral search as well, but it is up to you.

tsufz:
Hi Tomas,
Good to have fixed that. MZmine appears now in the log files...

PR #178, opened 2016-01-23
Two updates

tomas-pluskal:
* added reading zipped/gzipped raw data files
* chromatogram builder does not crash when no scans are selected, but instead quietly produces an empty peak list


tomas-pluskal:
@photocyte

PR #181, opened 2016-01-24
mzTab import - workaround to import .mzML.gz raw files

photocyte:


tomas-pluskal:
Looks good, Tim. Can you also add a check for .zip extension?

PR #182, opened 2016-01-24
ZIP import - Workaround for GZIP import GUI progress bar updating

photocyte:


src/main/java/net/sf/mzmine/modules/visualization/peaklisttable/PeakListTablePopupMenu.java
```diff
@@ -339,6 +339,16 @@ public void actionPerformed(final ActionEvent e) {
...
                         mzRange = peak.getRawDataPointsMZRange();
+                        double upper = mzRange.upperEndpoint();
+                        double lower = mzRange.lowerEndpoint();
+                        if ((upper - lower) < 0.000001) {
+                            //Workaround to make ultra narrow mzRanges (e.g. from imported mzTab peaklist), 
+                            //a more reasonable default for a HRAM instrument (~5ppm)
+              
+                            double fiveppm = (upper*5E-6);
+                            Range<Double> tempmzRange = Range.open(lower-fiveppm,upper+fiveppm);
+                            mzRange = tempmzRange;
```
tomas-pluskal: I think it is more intuitive to use Range.closed instead of Range.open. 
The tempmzRange variable is not necessary here, you can change it to a single line:
```
mzRange = Rangeclosed(lower-fiveppm,upper+fiveppm);
```

src/main/java/net/sf/mzmine/modules/rawdatamethods/rawdataimport/fileformats/ZipReadTask.java
```diff
@@ -189,6 +192,8 @@ public String getTaskDescription() {
     public void cancel() {
         if (copy != null)
             copy.cancel();
+        //Is it possible to delete the temporary files here?  
+        //Otherwise canceling the task potential leaves an incompletely decompressed file
```
tomas-pluskal: I will add that, although the files should still be deleted when the JVM exits, because they were marked by 
```
tmpFile.deleteOnExit();
tmpDir.deleteOnExit();
```
See https://docs.oracle.com/javase/7/docs/api/java/io/File.html#deleteOnExit()

photocyte: Yes that is what I observed, but if there is no quitting before trying to reload a file, an exception is thrown as the existing cached .mzML is incomplete

PR #185, opened 2016-01-24
Added better cancellation handling

tomas-pluskal:


tomas-pluskal:
@photocyte 
I added the deletion of the temporary files and proper handling of task cancellation.

PR #201, opened 2016-02-06
Fixed GoogleAnalyticsTracker trying to get screen size in headless mode

golgauth:
Google analytics features were trying to send graphical information (screen size) in a non-graphical environment (computing server) as well (leading to exception being thrown). 
**Added conditional statements.**

_WARNING: I am not a 100% sure about the way I fixed it. @dyrlund, you wrote that module (if I remember correctly): could you please cross examine this pull request changes. Thanks._


dyrlund:
The **utmsr** dosen't seem to be a required part of the url and I therefore think we can safely leave it out. I do however not know how this will show up in Google Analytics but if it causes any issues then we can always set the default to **utmsr=1x1**. [Source](https://developers.google.com/analytics/devguides/collection/gajs/gaTrackingTroubleshooting).

PR #203, opened 2016-02-12
CSVExportModule able to handle several peaklists as input

golgauth:


src/main/java/net/sf/mzmine/modules/peaklistmethods/io/csvexport/CSVExportTask.java
```diff
@@ -75,35 +78,53 @@ public double getFinishedPercentage() {
...
-        totalRows = peakList.getNumberOfRows();
+        // Process peak lists
+        for (PeakList peakList: peakLists) {
 
-        exportPeakList(peakList, writer);
+            // Filename
+            File curFile = fileName;
+            if (substitute) {
+                curFile = new File(fileName.getPath().replaceAll(
+                        Pattern.quote(plNamePattern), peakList.getName()));
```
tomas-pluskal: I think you should add some cleanup of the peak list name here.
peakList.getName() might return all kinds of weird characters like $, /, \, etc.

src/main/java/net/sf/mzmine/modules/peaklistmethods/io/csvexport/CSVExportTask.java
```diff
@@ -75,35 +78,53 @@ public double getFinishedPercentage() {
...
+                writer = new FileWriter(curFile);
+            } catch (Exception e) {
+                setStatus(TaskStatus.ERROR);
+                setErrorMessage("Could not open file " + curFile 
+                        + " for writing.");
+                return;
+            }
+
+            exportPeakList(peakList, writer, curFile);
+
```
tomas-pluskal: After exportPeakList() we should add a check if the task has been canceled
```
if (isCanceled()) return;
```
Otherwise the task will just continue with next peak list

src/main/java/net/sf/mzmine/modules/peaklistmethods/io/csvexport/CSVExportTask.java
golgauth: Yep, that would be better. And it makes me think that we should be doing this in the "DirectoryComponent", "FileNameComponent", etc. Because, if I am correct, if we type manually illegal characters in their `JTextField` (rather than using the file chooser), there is no check done there either.

tomas-pluskal:
Looks good!

golgauth:
Did you see this comment? What would you say?
https://github.com/mzmine/mzmine2/pull/203#discussion_r52716300

tomas-pluskal:
Sure, I agree it is a good idea to add the filename check to the other components as well.
Can you? :)

golgauth:
Hi Tomas,

OK, acting directly at the `Component` level (in the above code change I was only acting at the `Module` level) is doable since curly braces are not part of the forbidden file name characters. 

The forbidden characters are:
```
Linux (and Mac, I guess):
/ (forward slash)

Windows:
< (less than)
> (greater than)
: (colon)
" (double quote)
/ (forward slash)Component` level (in the above code change I was only acting at the `Module` level) is doable since curly braces are not part of the forbidden file name characters. 

\ (backslash)
| (vertical bar or pipe)
? (question mark)
* (asterisk)
```

So yes, this can be done. I'll probably add a modal dialog to warn the user when typing illegal characters. I'll also propagate the possibility to handle multiple input Peak lists to the other exporters. Those are `XMLExport`, `MZTabExport` and `SQLExport`. 

Did I miss anything, here?

Cheers

tomas-pluskal:
The SQLExport does not export to a file, so I think no need to make changes there. 
Otherwise, looks good.

PR #204, opened 2016-02-12
Fixed the peak shape component in case of MS2 (e.g, SRM) peaks

tomas-pluskal:


tomas-pluskal:
Also added fix for issue #205

PR #211, opened 2016-02-17
Row filter negation

photocyte:
Added the ability to remove peaklist rows that match all the conditions in the peaklist row filter (instead of keeping the rows based on the conditions).  Did do some tests to verify, but it is possible that I missed some edge cases with nulls etc.


photocyte:
screen shot 2016-02-17 at 12 46 08 pm

Picture for explanation

tomas-pluskal:
Looks good!

PR #214, opened 2016-02-17
Added a "Round resample filter" to the scan by scan filtering methods

golgauth:
See MZmine help page for more details.

Allows to:
* Readjust on exact ions values both "centroid" or "profile" spectra
* Option to cleanup zero intensity data points while processing
* Replaced the ancient "ScanUtils.detectSpectrumType" inner algorithm by the one from [msdk](https://github.com/msdk/msdk/blob/master/msdk-spectra/msdk-spectra-spectrumtypedetection /src/main/java/io/github/msdk/spectra/spectrumtypedetection/SpectrumTypeDetectionAlgorithm.java) (known to be more reliable)

__Note: In the case of "profile" spectra, each scan is converted to centroid using a local maxima detection algorithm first.__


tomas-pluskal:
Looks good to me.

PR #215, opened 2016-02-17
Module added: peak comparison row filter

photocyte:
This module filters peaklist rows based on comparisons between 2 peaks.  As of now a fold change comparison between two peaks of a given row is implemented.  Most of the code was copied from the rowsfilter module.


photocyte:
screen shot 2016-02-17 at 6 21 22 pm

Example given

src/main/java/net/sf/mzmine/modules/peaklistmethods/filtering/peakcomparisonrowfilter/PeakComparisonRowFilterTask.java
```diff
@@ -0,0 +1,318 @@
...
+
+            final Feature newPeak = new SimpleFeature(peak);
+            PeakUtils.copyPeakProperties(peak, newPeak);
+            newRow.addPeak(peak.getDataFile(), newPeak);
+        }
+
+        return newRow;
+    }
+
+    private int getPeakCount(PeakListRow row, String groupingParameter) {
```
tomas-pluskal: Is the getPeakCount() method ever used? 
If not, you can remove it..

photocyte:
Changes made

PR #219, opened 2016-02-24
Add support for PeakInvestigator (MassDetection)

TDI-tenderholt:
This pull request implements the PeakInvestigator public API, which is used for a new MassDetector module. See http://veritomyx.com and https://peakinvestigator.veritomyx.com/api/ for more information.


tomas-pluskal:
Hi Adam,
I am afraid it is impossible for us to review 250+ commits that make changes all over the MZmine code base. Can we try a better way to do this?

My suggestion is:
1) Start with a fresh clone of the latest MZmine 2 master branch
2) In a single commit, add all classes related to your new MassDetector. Please add them all into the package designated for this mass detector, do not add new packages like com.veritomyx... If your code requires some common utility classes that are used by your company, you can add them as a new jar dependency.
3) In another commit, please add the changes to existing MZmine classes. The changes to existing code should be done systematically, with comments, and for good reasons. As a bad example, see your proposed changes to .gitignore or MassDetector.java in this pull request. Especially, the changes made to MassDetector.java are very confusing.
4) In another commit, please add any supporting files like libraries or icons

That way we can review the changes and merge them.

TDI-tenderholt:
Thanks for the feedback—I had thought that it might be too crazy to review all the changes, but figured I'd go ahead and create the PR as you suggested. And unfortunately, there have been two other developers over the years so I think cleaning up everything will take several days.

A few questions:

* Do you not want the code to compile after each commit? Simply adding classes for the new PeakInvestigatorDetector, which depend on changes to other classes, will not compile. Similarly, adding libraries last will lead to compile errors.
* Can you provide comment on additions to ```mzmine.util``` (https://github.com/Veritomyx/mzmine2/tree/master/src/main/java/net/sf/mzmine/util/dialogs)? I created these dialogs & factories to make it easy to write unit tests that don't depend on any GUI.
* I believe changes to ```MassDetector.java``` were made to facilitate how calls to PeakInvestigator work. Specifically, there needs to be a step to upload data and another step to download data. Are modifications made to this class allowed or will you reject any code that requires modifying this class?
* What is the timeframe for MSDK and MZmine3?

tomas-pluskal:
It does not necessarily need to compile after each commit, but I guess we could start with the most "harmless" changes, like adding libraries and icons, first?

The Dialog classes I don't like so much - because they're only used by your module. But you can keep them and move them to your module's package.

Changes to MassDetector.java are OK but should be more systematic. Currently, the MassDetector has a single method, getMassValues(). You can add methods like initMethod() and finishMethod(), but I don't see the point of adding so many parameters.

My timeframe for MZmine 3 is to have it released this year. The current development version is 100% based on MSDK, and MSDK is already quite stable and usable for many things. We publish MSDK releases to Maven Central: http://search.maven.org/#search%7Cga%7C1%7Cg%3A%22io.github.msdk%22

tomas-pluskal:
Hi Adam,
FYI, I am planning to release MZmine 2.20 in a few days. But we can leave the PeakInvestigator integration for next version, if you're not in a rush to have it in the "official" MZmine.

TDI-tenderholt:
Thanks for the info. We're not in a rush to have it in the official MZmine.

Out of curiosity, what changed between 2.19 and 2.20? Is there a time frame for 2.21?

tomas-pluskal:
Here you can see the commits since v. 2.19:

https://github.com/mzmine/mzmine2/compare/v2.19...master

There is no timeframe. We can release 2.21 after the PeakInvestigator integration is complete.

PR #225, opened 2016-03-05
peak comparison filter - null handling conditional was missing brackets

photocyte:
Sorry missed this one in testing (test peaklist had no nulls)


tomas-pluskal:
Thanks!

PR #227, opened 2016-03-14
docs: fix typos in three documentation files

Tebatu:
I fixed two typos and one missing special character in the help files.


tomas-pluskal:
Thanks a lot!

PR #230, opened 2016-03-30
FileNameParameter: ignore case while appending extension

golgauth:
Windows mostly users can have files with uppercase extension.
This fix avoids to get a file named `/dir/file.EXT` to get selected (and fail to be selected) as `/dir/file.EXT.ext` in case FileNameParameter is build with `'.ext'` as extension argument.

### DIFF
--- a/src/main/java/net/sf/mzmine/parameters/parametertypes/filenames/FileNameParameter.java
+++ b/src/main/java/net/sf/mzmine/parameters/parametertypes/filenames/FileNameParameter.java
@@ -99,9 +99,8 @@ public class FileNameParameter
     public void setValueFromComponent(FileNameComponent component) {
         File compValue = component.getValue();
         if (extension != null) {
-            if (!(compValue.getName().toUpperCase()
-                    .endsWith(extension.toUpperCase()) || compValue.getName()
-                    .toLowerCase().endsWith(extension.toLowerCase())))
+            if (!compValue.getName()
+                    .toLowerCase().endsWith(extension.toLowerCase()))
                 compValue = new File(compValue.getPath() + "." + extension);
         }
         this.value = compValue;
--- a/src/main/java/net/sf/mzmine/parameters/parametertypes/filenames/FileNameParameter.java
+++ b/src/main/java/net/sf/mzmine/parameters/parametertypes/filenames/FileNameParameter.java
@@ -99,7 +99,9 @@ public class FileNameParameter
     public void setValueFromComponent(FileNameComponent component) {
         File compValue = component.getValue();
         if (extension != null) {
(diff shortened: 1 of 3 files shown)


src/main/java/net/sf/mzmine/parameters/parametertypes/filenames/FileNameParameter.java
```diff
@@ -99,7 +99,9 @@ public FileNameParameter cloneParameter() {
     public void setValueFromComponent(FileNameComponent component) {
         File compValue = component.getValue();
         if (extension != null) {
-            if (!compValue.getName().endsWith(extension))
+            if (!(compValue.getName().toUpperCase()
+                    .endsWith(extension.toUpperCase()) || compValue.getName()
+                    .toLowerCase().endsWith(extension.toLowerCase())))
```
tomas-pluskal: It's kinda unnecessary to test both toUpperCase() and toLowerCase() - think about it :)

golgauth: Yep, you are perfectly right ;)

I meant:

           if (!(compValue.getName().toUpperCase()
                   .endsWith(extension) || compValue.getName()
                   .toLowerCase().endsWith(extension)))

The idea being to prevent cases where case is mixed up like ".JpeG", ...

I'll fix it that way.

tomas-pluskal: No, I think you meant:
````
  if (!(compValue.getName().toUpperCase().endsWith(extension.toUpperCase()))
````

golgauth: Well, no. I did not. Here is what I want:

* "file.EXT" remains "file.EXT"
* "file.ext" remains "file.ext"
* "file.ExT" becomes "file.ExT.ext"

Because extension case shouldn't be mixed up, the latter requires the extension be: `not("EXT" or "ext")` to be appended. I feel like I never wrote a line of code if the second proposition is not correct ;) ...

But maybe you find it useless to cover the third case, which is an argument I can eventually share (in which case your line is correct).

tomas-pluskal: I see. Well actually I do find the the third case a bit awkward. I would prefer to just check for the existence of the extension, and ignore the case completely.

golgauth: Yep I agree, this 3rd case is usually not covered in other applications (that was a digression of mine). Cheers ;)

PR #246, opened 2016-06-07
Ms2 similarity score identification module

photocyte:


src/main/java/net/sf/mzmine/modules/peaklistmethods/identification/ms2search/Ms2SearchTask.java
```diff
@@ -0,0 +1,280 @@
...
+                + peakList2);
+
+    }
+
+    /**
+     * Simple cosine like calculation of similarity between ions of two fragmentation scans.
+     * 
+     */
+    private Ms2SearchResult simpleMS2similarity(Feature featureA, Feature featureB,
+            double intensityThreshold, double mzRangePPM) {
```
tomas-pluskal: This function should take MZTolerance object instead of double as the mzRangePPM parameter.

src/main/java/net/sf/mzmine/modules/peaklistmethods/identification/ms2search/Ms2SearchTask.java
```diff
@@ -0,0 +1,280 @@
...
+                Feature featureB = rows2[j].getBestPeak();
+                
+                int featureAID = featureA.hashCode();
+                int featureBID = featureB.hashCode();
+                
+                searchResult = simpleMS2similarity(featureA,
+                        featureB, intensityThreshold, mzTolerance.getPpmTolerance());
+                
+                //Report the final score to the peaklist identity
+                if (searchResult != null && searchResult.getScore() > scoreThreshold && searchResult.getNumIonsMatched() >= minimumIonsMatched)
```
tomas-pluskal: In java it is a good practice to avoid many function calls on a single line - if an exception is thrown, you will not know what exactly happened, because the JVM will only report the line number where the exception occurred.

src/main/java/net/sf/mzmine/modules/peaklistmethods/identification/ms2search/Ms2Identity.java
```diff
@@ -0,0 +1,39 @@
...
+import net.sf.mzmine.datamodel.Feature;
+import net.sf.mzmine.datamodel.PeakListRow;
+import net.sf.mzmine.datamodel.impl.SimplePeakIdentity;
+import net.sf.mzmine.main.MZmineCore;
+
+public class Ms2Identity extends SimplePeakIdentity {
+
+    public Ms2Identity(final Feature featureA, final Feature featureB, Ms2SearchResult searchResult) {
+
+        super("MS2 similarity m/z:" + MZmineCore.getConfiguration().getMZFormat().format(featureB.getMZ()) +
```
tomas-pluskal: Here also, it would be better to avoid so many function calls on a single line. Splitting this to several lines is better for readability. You don't have to change it now, just for next time.

src/main/java/net/sf/mzmine/modules/peaklistmethods/identification/ms2search/Ms2SearchTask.java
```diff
@@ -0,0 +1,280 @@
...
+        int rows2Length = rows2.length;
+        
+        totalRows = rows1Length;
+        
+        for (int i = 0; i < rows1Length; i++) {
+            for (int j = 0; j < rows2Length; j++) {
+                Feature featureA = rows1[i].getBestPeak();
+                Feature featureB = rows2[j].getBestPeak();
+                
+                int featureAID = featureA.hashCode();
```
tomas-pluskal: hashCode()?

src/main/java/net/sf/mzmine/modules/peaklistmethods/identification/ms2search/Ms2SearchTask.java
```diff
@@ -0,0 +1,280 @@
...
+        
+        if (scanMS2A == null || scanMS2B == null)
+        {
+            return null;
+        }
+
+        DataPoint[] ionsA = null;
+        DataPoint[] ionsB = null;
+
+        MassList massListA = scanMS2A.getMassLists()[0];
```
tomas-pluskal: scanMS2A.getMassLists()[0] will throw an exception if there is no mass list at all. Plus, taking the first mass list might not always be optimal - why not add the MassListParameter to the parameter set and let the user choose which mass list to use.

src/main/java/net/sf/mzmine/modules/peaklistmethods/identification/ms2search/Ms2SearchTask.java
photocyte: Was attempting to get a unique identifier per feature, as in the future it may be better to store exactly what feature was searched against & what the results were.  This gets more complicated with the multiple features per single row in the feature list.

src/main/java/net/sf/mzmine/modules/peaklistmethods/identification/ms2search/Ms2SearchTask.java
```diff
@@ -0,0 +1,280 @@
...
+        if (ionsA == null || ionsB == null)
+        {
+            //Fall back to profile data?
+             //Profile / raw data.
+            //ionsA = scanMS2A.getDataPointsOverIntensity(intensityThreshold);
+            //ionsB = scanMS2B.getDataPointsOverIntensity(intensityThreshold);
+         return null;   
+        }
+        
+        // Compare every ion peak in MS2 scan A, to MS2 scan B.  Possible heuristics to speed this up based on sorted mz values etc.
```
tomas-pluskal: The data points are always sorted by m/z, so these two cycles can be optimized.

src/main/java/net/sf/mzmine/modules/peaklistmethods/identification/ms2search/Ms2SearchTask.java
photocyte: I see, yes that can be added.

src/main/java/net/sf/mzmine/modules/peaklistmethods/identification/ms2search/Ms2SearchTask.java
```diff
@@ -0,0 +1,280 @@
...
+        logger.info("Finished MS2 similarity search for " + peakList1 + "against"
+                + peakList2);
+
+    }
+
+    /**
+     * Simple cosine like calculation of similarity between ions of two fragmentation scans.
+     * 
+     */
+    private Ms2SearchResult simpleMS2similarity(Feature featureA, Feature featureB,
```
tomas-pluskal: I would suggest to move this function into a separate class (e.g. CosineProductSimilarity) and move it to the util package. Then we can use the function here as well as in the scatter plot, for example.

PR #278, opened 2017-03-08
ADAP incorporated into MZmine 2 

du-lab:
Dear MZmine 2 development team,

My Du-lab team have incorporated two new modules we developed into MZmine 2. One module carries out chromatogram building and the other module carries out chromatogram deconvolution. These two modules can be used for both LC-MS and GC-MS data processing.

In addition to these additions to MZmine 2, we have also made some changes to MZmine 2 itself. Please see page 36 of the attached ADAP User Manual for the list of major changes.

There are also some minor changes we made to MZmine 2 that are not listed on this page.

In addition to the list of major changes, this pdf attachment also includes a step-by-step procedure about how to use the two new modules in the context of MZmine 2. 

Please ignore Chapter 4 "ADAP-GC" for now.

[ADAP_user_manual.pdf](https://github.com/mzmine/mzmine2/files/828729/ADAP_user_manual.pdf)

Thank you very much for reviewing the additions and changes. Please let me know if you have any questions.

Best regards,
Xiuxia Du
xiuxia.du@uncc.edu
704-687-7307

PR #281, opened 2017-03-30
addressed issues related to pull request #278

du-lab:
Hi Tomas,

We have addressed the compile errors you pointed out. In addition, the Wavelets ADAP (GC) module is removed from the chromatogram deconvolution since we have replaced it with a different module.

Please review the changes and let me know any questions you might have.

Thank you for your time and consideration.

Best regards,
Xiuxia on behalf of the Du-Lab team


tomas-pluskal:
Thank you, looks good!

Issue #282, opened 2017-03-30
ADAP chromatogram builder cancel()


tomas-pluskal:
It seems the cancel() method is not working in the ADAP chromatogram builder module.

tomas-pluskal:
@du-lab

PR #283, opened 2017-03-31
fixed cancel() in ADAP chrom builder

du-lab:
Hi Tomas,

I am making a PR after we have fixed the issue you pointed out in

https://github.com/mzmine/mzmine2/issues/282

Please review and let me know if you have any questions for us.

Xiuxia on behalf of the Du-Lab team


tomas-pluskal:
Thanks!

PR #301, opened 2017-05-03
Adding modules for MGF export of LC-MS/MS data for GNPS and Sirius, and a new filter to keep only features with MS/MS

lfnothias:
The DorresteinLab "bioinformatic squad" prepared MZmine modules to export LC-MS/MS features as MGF that can be processed with [GNPS](http://gnps.ucsd.edu), and with [Sirius](https://bio.informatik.uni-jena.de/software/sirius/).
In addition, we created a new filter that keep only features with MS/MS spectrum associated.

Credit to Prof. Xiuxiua Du' group, as we benefited from her MGF export module. And especially to  new Dorrestein lab member's at University of California San Diego: Abi Sarvepalli, Christopher Jensen, and Zheng Zhang. The effort was coordinated by myself and Dr. Daniel Petras.

Cheers,

Louis-Felix


tomas-pluskal:
Thanks, guys! It looks good.

One question: what is the purpose of making a new PeakFilterMS2Module? Wouldn't it make more sense to add this function to the existing peak list filtering module?

lfnothias:
Cool !

We were not sure which one was the best approach for implementation. If you
advise for that choice, lets keep only the function that is present in the
peak list filtering module. We are going to commit required changes right
now.

LF

tomas-pluskal:
I think having multiple modules that do essentially the same is confusing for the users, so it's better to integrate it into the existing module.

lfnothias:
Tomas, 
Zheng just removed the duplicate module. Let us know if anything else must be changed.

LF

tomas-pluskal:
One more question: why do you convert the m/z values to integers for export to Sirius (https://github.com/mzmine/mzmine2/pull/301/files#diff-8609bf3190b03e1dacd75019e616489aR222)? I thought Sirius can take advantage of accurate m/z values for formula prediction.

lfnothias:
Correct. Thanks for seing that. 
Let us fix it.

LF

lfnothias:
We kept the fractional m/z function as comment, as it will certainly be useful for other tools currently in development.
One question about the association between one feature and its "bestMSMSscan". I guess there a retention time and m/z tolerance for that pairing ? Could you explain where these parameters are defined ?

tomas-pluskal:
The bestMSMSscan is defined during peak detection. For each peak, the highest intensity MS/MS scan is selected from all MS/MS scans with the right precursor m/z and RT within the RT span of the peak.

lfnothias:
Ok, just to make sure I got it right, when you say right precursor, it means that match both the m/z and RT max ranges at peak detection step ?

tomas-pluskal:
Yup

lfnothias:
Great. It will work for perfectly for Bruker qTOF and Thermo QE data. However, it will not work for Agilent qTOF data, as the precursor ion value is not recorded accurately (like 300.1 Da). It is really BS. Because using larger m/z window might impact overall peak detection. So we should probably consider having a specific parameters for that m/z window to get the bestMSMSscan. Could you comment on that possibility and we can help implementing that if needed.
LF

tomas-pluskal:
Sure, I can help you with that. Perhaps you should do some testing on Agilent data first, and then let's discuss.
I am going to merge this PR now.

lfnothias:
Alright and thanks !

lfnothias:
Hello Tomas, we get some feedback and we found couple of modifications that are needed to allow smooth processing of the output files with GNPS (using row filter instead of peak filter) and Sirius (precursor ion header in the MGF file for the MS1 scan). We are committing and will prepare the pull request very soon.

PR #303, opened 2017-05-10
Update of GNPS/Sirius export module + new filter for peak ID reset

lfnothias:
Dear Tomas,
We fixed couple of elements that prevented smooth processing of the files. Including:
 - Small changes for the GNPS/Sirius export modules.
- Peak ID reset option.
Note that these module GNPS/Sirius are not functional until the new release.
Let me know if anything needs to be fixed.

LF


tomas-pluskal:
Looks good, thanks.

lfnothias:
Ok. I double-checked the data processing with both GNPS and Sirius and it works perfectly now. 
Could you let us know when will be next release ? Many lab members are excited here about trying the module :)

tomas-pluskal:
I can do that tonight.

lfnothias:
That would be awesome and greatly appreciated. Thank you very much.

PR #307, opened 2017-05-30
Total memory detection in other locales

zeehio:
When `free` is localized the `awk` line fails to parse the total memory, for instance on Ubuntu Trusty under a Spanish locale.

This commit ensures that `free` is run with the C locale, so the output is in English, as the awk script expects.


tomas-pluskal:
Looks good, thanks!

PR #317, opened 2017-06-19
DorresteinLab Chromatogram Pairing PR 

abisarvepalli:


tomas-pluskal:
Hi,
Looks much better now!
I see the PR contains a change to src/main/lib/dulab/adap/3.2.0/adap-3.2.0.jar, please check that.

tomas-pluskal:
Also, the Travis build check failed. Please take a look at the error log.

abisarvepalli:
Checking right now!

abisarvepalli:
Just fixed the .jar file

src/main/java/net/sf/mzmine/datamodel/impl/SimplePeakIdentity.java
```diff
@@ -131,9 +131,9 @@ public String toString() {
     }
 
     @Override
-    // Removed @Nonnull in front because the function may return null
-    public String getPropertyValue(final String property) {
-		return properties.get(property);
+    public @Nonnull String getPropertyValue(final String property) {
```
tomas-pluskal: This change should not be here

src/main/java/net/sf/mzmine/modules/masslistmethods/chromatogrambuilder/ChromatogramBuilderTask.java
```diff
@@ -191,6 +211,7 @@ public void run() {
         }
 
         // Add new peaklist to the project
+        System.out.print(this.RTRangeMSMS);
```
tomas-pluskal: pls use the logger here

tomas-pluskal:
A question: you added two new parameters to chromatogram builder, but I think these parameters would actually make more sense in the chromatogram deconvolution module. Because that is the module, which ultimately decides the MS/MS spectrum for a peak. And you also use the saved values from chromatogram builder in there - however, it won't work if the user uses the ADAP chromatogram builder, for example.

tomas-pluskal:
Adding the getMZrangeMSMS and getRTrangeMSMS methods to all Feature implementing classes seems unnecessary - you wouldn't need that if you added the parameters to the chromatogram deconvolution module.

src/main/java/net/sf/mzmine/modules/rawdatamethods/peakpicking/manual/ManualPeak.java
```diff
@@ -309,4 +309,16 @@ public SimplePeakInformation getPeakInformation(){
     }
     //End dulab Edit
 
+	@Override
```
tomas-pluskal: I guess these edits to ManualPeak are not needed anymore?

abisarvepalli: Ya getting rid of that right now. It was the reason for the syntax errors that caused errors in the checks.

abisarvepalli:
With the omission of the override tags, all checks have passed and think that this is ready!

tomas-pluskal:
The PR contains some changes to adap3decompositionV1_5/help/help.html
Did you make those changes, or are you reverting somebody else's commit?

abisarvepalli:
Seems that we were reverting somebody else's commit and that issue has been cleared.

tomas-pluskal:
OK, looks good.

lfnothias:
It's good to hear ! Now it's good to go on our side. Thank you very much for your help and your patience.

Issue #298, opened 2017-05-02
xcms integration: chromatogram deconvolution fails with recent version of xcms


rababerladuseladim:
Windows Version: 10 x64
java: v8 update 121
R: 3.4.0
mzmine: 2.24
xcms: 1.52.0

Hi, 
we recently updated our R installation and since the update we cannot deconvolute our chromatograms with the wavelet mthod of xcms anymore.

The associated error message:

`[16:16:27|FINEST|MainMenu]: Setting parameters for module Chromatogram deconvolution
[16:16:49|FINEST|PeakResolverSetupDialog]: Loading new preview peak #1 84.960 m/z @20.5
[16:16:49|FINEST|RLocationDetection]: Windows: Query registry to find where R is installed ...
[16:16:49|FINEST|RLocationDetection]: R InstallPath = C:\Program Files\R\R-3.4.0
[16:16:49|FINEST|RLocationDetection]: R executable location auto-detected as 'C:\Program Files\R\R-3.4.0\bin\R.exe'
[16:16:50|FINEST|RSessionWrapper]:   failed: timed out
[16:16:50|FINEST|RSessionWrapper]: Connection R://MZmineUser:8d4e3598-6651-406f-b430-2c9689042819@localhost:6311 failed.
[16:16:50|FINEST|RSessionWrapper]: !! Rserve R://MZmineUser:8d4e3598-6651-406f-b430-2c9689042819@localhost:6311 is not accessible.
 ! null
[16:16:50|FINEST|RSessionWrapper]: Trying to spawn R://MZmineUser:8d4e3598-6651-406f-b430-2c9689042819@localhost:6311
[16:16:50|FINEST|RSessionWrapper]: Failed to connect: Cannot connect: Connection refused: connect
[16:16:50|FINEST|RLocationDetection]: Windows: Query registry to find where R is installed ...
[16:16:50|FINEST|RLocationDetection]: R InstallPath = C:\Program Files\R\R-3.4.0
[16:16:50|FINEST|RLocationDetection]: R executable location auto-detected as 'C:\Program Files\R\R-3.4.0\bin\R.exe'
[16:16:50|FINEST|RSessionWrapper]: checking Rserve is available... 
[16:16:50|FINEST|RSessionWrapper]:   executing "C:\Program Files\R\R-3.4.0\bin\R.exe" -e "i=installed.packages();is.element(set=i,el='Rserve')" --vanilla -q
[16:16:50|FINEST|RSessionWrapper]:   ok
[...]

PR #321, opened 2017-07-04
'Fixed chromatogram deconvolution fails with recent version of xcms #298

golgauth:
Fix should be compatible with older R releases since I think this was caused by an additional check in R 3.4.X.

Please tell me if this work as expected with this fix in _centWave_ R code.


tomas-pluskal:
Thanks a lot, Gauthier!

PR #327, opened 2017-07-11
Add GNPS&SIRIUS Module Disclaimers

abisarvepalli:
# Added GNPS Disclaimer to SetupDialog
# Added Sirius Disclaimer to SetupDialog


src/main/java/net/sf/mzmine/modules/peaklistmethods/io/gnpsexport/GNPSSetupDialog.java
```diff
@@ -0,0 +1,325 @@
...
+    private JButton btnOK, btnCancel;
+
+	
+	
+	protected GridBagPanel mainPanel;
+
+    /**
+     * Constructor
+     */
+    public GNPSSetupDialog(Window parent, boolean valueCheckRequired, ParameterSet parameters) {
```
tomas-pluskal: Instead of making a copy of the ParameterSetupDialog class, why don't you extend it?

src/main/java/net/sf/mzmine/modules/peaklistmethods/io/siriusexport/SiriusExportParameters.java
```diff
@@ -42,13 +47,19 @@
     public static final ComboParameter <String> ROUND_MODE = 
             new ComboParameter <> (
                     "Merging Mode", 
-                    "Determines how to merge intensities with the same m/z values",
+                    "Detexrmines how to merge intensities with the same m/z values",
```
tomas-pluskal: Detexrmines?

src/main/java/net/sf/mzmine/modules/peaklistmethods/io/gnpsexport/GNPSSetupDialog.java
tomas-pluskal: In fact, it would be best if you could add a support for this function to ParameterSetupDialog, so other modules can use it as well.

abisarvepalli: The issue with extending the class is that the parameters are protected and the addDialogComponents() needs to be overridden by the new class. Provided that most of the parameters are protected, I thought it was best to clone the class for now and then provide a better implementation by the next update.

tomas-pluskal: Yeah, I know that trick... next update is never going to happen. Please provide a better implementation now. Feel free to change the visibility of some fields in the ParameterSetupDialog if necessary (if the fields are declared protected they should be accessible to derived classes!). Also you can override the addDialogComponents() method, call super.addDialogComponents() first and then add your additions.

abisarvepalli:
Hey Tomas,

Could you please check if this implementation works better?

tomas-pluskal:
I'd say it's almost perfect. Why not add the footer functionality to the ParameterSetupDialog itself, instead of creating a new dialog class? If the footer message is null, just don't add the label.

abisarvepalli:
Sounds good

src/main/java/net/sf/mzmine/modules/peaklistmethods/io/siriusexport/SiriusExportParameters.java
```diff
@@ -46,9 +52,21 @@
...
     
-    public SiriusExportParameters() {
-	super(new Parameter[] {PEAK_LISTS, FILENAME, /*FRACTIONAL_MZ,*/ ROUND_MODE, MASS_LIST});
+    public ExitCode showSetupDialog(Window parent, boolean valueCheckRequired) { 
+    	String message = "<html>SIRIUS Module Disclaimer:" +
+        		"<br>    - If you use the SIRIUS export module, cite MZmine2 paper and the following articles: Dührkop et al.," +
+        		"<br>     Proc Natl Acad Sci USA 112(41):12580-12585 and Boecker et al., Journal of Cheminformatics (2016) 8:5" + 
+        		"<br>    - Sirius can be downloaded at the following address: <a href\"\">https://bio.informatik.uni-jena.de/software/sirius/</a>" +
+        		"<br>    - Sirius results can be mapped into GNPS molecular networks. See the documentation: "+
+        		"<br>     <a href=\"\">https://bix-lab.ucsd.edu/display/Public/GNPS+data+analysis+workflow+2.0</a>.</html>";
```
tomas-pluskal: Why is the link pointing to nowhere? (a href="")

src/main/java/net/sf/mzmine/modules/peaklistmethods/io/gnpsexport/GNPSExportParameters.java
```diff
@@ -36,4 +41,15 @@
...
     	super(new Parameter[] {PEAK_LISTS, FILENAME, MASS_LIST});
     }
-}
+    
+    public ExitCode showSetupDialog(Window parent, boolean valueCheckRequired) { 
+    	String message = "<html>GNPS Module Disclaimer:" + 
+        		"<br>    - If you use the GNPS export module for GNPS web-platform, cite MZmine2 paper and the following article:"+
+        		"<br>     Wang et al., Nature Biotechnology 34.8 (2016): 828-837." +
+        		"<br>    - See the documentation about MZmine2 data pre-processing for GNPS (<a href=\"\">http://gnps.ucsd.edu/</a>) molecular " +
+        		"<br>     networking and MS/MS spectral library search. <a href=\"\">https://bix-lab.ucsd.edu/display/Public/GNPS+data+analysis+workflow+2.0</a></html>";
```
tomas-pluskal: This link also pointing to nowhere?

src/main/java/net/sf/mzmine/modules/peaklistmethods/io/siriusexport/SiriusExportParameters.java
```diff
@@ -46,9 +52,21 @@
...
+    
+    public SiriusExportParameters() {    	
+    	super(new Parameter[] {PEAK_LISTS, FILENAME, ROUND_MODE, MASS_LIST});
+    }
     
-    public SiriusExportParameters() {
-	super(new Parameter[] {PEAK_LISTS, FILENAME, /*FRACTIONAL_MZ,*/ ROUND_MODE, MASS_LIST});
+    public ExitCode showSetupDialog(Window parent, boolean valueCheckRequired) { 
+    	String message = "<html>SIRIUS Module Disclaimer:" +
+        		"<br>    - If you use the SIRIUS export module, cite MZmine2 paper and the following articles: Dührkop et al.," +
```
tomas-pluskal: I think asking people to cite 2 papers about SIRIUS is a bit too much. I would recommend referencing only the PNAS paper.

src/main/java/net/sf/mzmine/modules/peaklistmethods/io/gnpsexport/GNPSExportParameters.java
abisarvepalli: The reason is because the html-tag provides the visual of a link, but regardless of whether or not href is populated, the label does not implement a listener to redirect to a website.

tomas-pluskal:
Almost there.
Can you change the first ParameterSetupDialog constructor to this:
````
public ParameterSetupDialog(Window parent, boolean valueCheckRequired,
            ParameterSet parameters) {
    this(parent, valueCheckRequired, parameters, null);
}
````
To avoid duplicating code. Duplicated code is evil.

tomas-pluskal:
Thanks!

tomas-pluskal:
I tested the new ParameterSetupDialog on my system. It doesn't look great:
screen shot 2017-07-16 at 22 37 45
I think it might be better to use JEditorPane instead of JLabel. That would also allow you to make the links functional, see https://coderanch.com/t/345547/java/Making-clickable-links-HTML-JLabel

abisarvepalli:
I will check out JEditorPane right now.

Thanks
Abi Sarvepalli

On July 16, 2017 at 7:41:57 PM, Tomáš Pluskal (notifications@github.com)
wrote:

I tested the new ParameterSetupDialog on my system. It doesn't look great:
[image: screen shot 2017-07-16 at 22 37 45]
<https://user-images.githubusercontent.com/1796692/28254079-81590a78-6a77-11e7-92f8-8a32125dadca.png>
I think it might be better to use JEditorPane instead of JLabel. That would
also allow you to make the links functional, see
https://coderanch.com/t/345547/java/Making-clickable-links-HTML-JLabel

—
You are receiving this because you authored the thread.
Reply to this email directly, view it on GitHub
<https://github.com/mzmine/mzmine2/pull/327#issuecomment-315658366>, or mute
the thread

src/main/java/net/sf/mzmine/modules/peaklistmethods/io/siriusexport/SiriusExportParameters.java
lfnothias: Actually, both citations are needed. Indeed Sirius 3.5 contains, Sirius algorithm (molecular formula from fragmentation tree), and CSIFingerID tool for structure search.

tomas-pluskal:
The GNPS paper link is pointing to a wrong URL:
https://www.nature.com/nbt/journal/v34/n8/full/nbt.3597.htm

PR #329, opened 2017-07-17
Footer Parameters for SetupDialog

abisarvepalli:
# Updated for GNPSExport
# Updated for SiriusExport

### DIFF
--- a/src/main/java/net/sf/mzmine/modules/peaklistmethods/io/gnpsexport/GNPSExportParameters.java
+++ b/src/main/java/net/sf/mzmine/modules/peaklistmethods/io/gnpsexport/GNPSExportParameters.java
@@ -14,7 +14,6 @@ package net.sf.mzmine.modules.peaklistmethods.io.gnpsexport;
 
 import java.awt.Window;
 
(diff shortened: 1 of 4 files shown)


tomas-pluskal:
The GNPS paper link is pointing to a wrong URL:
https://www.nature.com/nbt/journal/v34/n8/full/nbt.3597.htm

tomas-pluskal:
Also, I wonder if you're planning to add the help files?

lfnothias:
Thanks for seing that ! The l of html is missing

lfnothias:
We would rather keep the link to the documentation on the CCMS webpage in
the footer for now, since it will be the main gateway to GNPS and many
small update will be needed in the coming weeks.
However, we can duplicate the Footer in the Help documentation if you think
this is appropriate.

Regarding Sirius, I will ask Kai if they can coordinate the Help document.
But it will take some time.

LF

tomas-pluskal:
It would be great if you could add a help file for both modules, with a short (1 paragraph) explanation of what the module does and what the parameters mean. Just copy/paste a template from other modules.

lfnothias:
Alright. We are on it.

PR #332, opened 2017-07-28
Update MS/MS filtering, GNPS/Sirius export

lfnothias:
-  MS/MS filtering fix.
- Update GNPS/Sirius export module: MS2 with higher base peak intensity across aligned peaks is now exported. 
- Typo correction on footer.


tomas-pluskal:
The PR changes permissions on RowsFilterTask.java, could you please revert that change?

PR #349, opened 2017-08-10
Update GNPSExportTask

abisarvepalli:
- Added support for future module functionality by exporting blank
scans for scans lacking MS2 scans.

### DIFF
--- a/src/main/java/net/sf/mzmine/modules/peaklistmethods/io/gnpsexport/GNPSExportTask.java
+++ b/src/main/java/net/sf/mzmine/modules/peaklistmethods/io/gnpsexport/GNPSExportTask.java
@@ -32,8 +32,6 @@ import net.sf.mzmine.parameters.ParameterSet;
 import net.sf.mzmine.taskcontrol.AbstractTask;
 import net.sf.mzmine.taskcontrol.TaskStatus;
 import net.sf.mzmine.util.PeakUtils;
-import weka.core.logging.Logger;
-import weka.core.logging.Logger.Level;
 
 public class GNPSExportTask extends AbstractTask {
   private final PeakList[] peakLists;
--- a/src/main/java/net/sf/mzmine/modules/peaklistmethods/io/gnpsexport/GNPSExportTask.java
+++ b/src/main/java/net/sf/mzmine/modules/peaklistmethods/io/gnpsexport/GNPSExportTask.java
@@ -32,272 +32,240 @@ import net.sf.mzmine.parameters.ParameterSet;
 import net.sf.mzmine.taskcontrol.AbstractTask;
 import net.sf.mzmine.taskcontrol.TaskStatus;
 import net.sf.mzmine.util.PeakUtils;
+import weka.core.logging.Logger;
+import weka.core.logging.Logger.Level;
 
 public class GNPSExportTask extends AbstractTask {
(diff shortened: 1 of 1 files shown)


src/main/java/net/sf/mzmine/modules/peaklistmethods/io/gnpsexport/GNPSExportTask.java
```diff
@@ -32,272 +32,240 @@
 import net.sf.mzmine.taskcontrol.AbstractTask;
 import net.sf.mzmine.taskcontrol.TaskStatus;
 import net.sf.mzmine.util.PeakUtils;
+import weka.core.logging.Logger;
```
tomas-pluskal: importing logging from Weka?

src/main/java/net/sf/mzmine/modules/peaklistmethods/io/gnpsexport/GNPSExportTask.java
```diff
@@ -32,272 +32,240 @@
...
-                setStatus(TaskStatus.ERROR);
-                setErrorMessage("Error while writing into file " + curFile +
-                        ": " + e.getMessage());
-                return;
-            }
+    int currentID = 1;
+    for (int i = 0; i < peakList.getRows().length; i++) {
+      PeakListRow row = peakList.getRows()[i];
+      String rowID = Integer.toString(row.getID());
+      if (currentID != row.getID()) {
```
tomas-pluskal: This does not make sense to me - there is no guarantee that the IDs  start from 1 and are increasing. It depends on how the peaklist was produced.

abisarvepalli: The point of this functionality is to provide filler output for scans starting with scan #1. 
We traverse the array list of peakLists and, for the missing scan values, write the corresponding scan number to the output file.

tomas-pluskal: Peak list row is a different thing from scan. 
Why would anyone want to you want export non-existing rows?

abisarvepalli: It is intended to comply with the GNPS Molecular Network format.

tomas-pluskal: Why would GNPS want to import non-existing data?

abisarvepalli: These are necessary for existing file readers and backwards compatibility in gnps

tomas-pluskal: So, if I have a peak list with a single row, the ID of which is 1000, you are going to export 999 empty lines in front of it, because that's what the GNPS parser requires? 
Wouldn't it be better to fix the GNPS parser instead of adding this workaround to MZmine?

tomas-pluskal:
Could you please also update the getFinishedPercentage() method to return 1.0 when the task is finished?

PR #353, opened 2017-08-17
Update GNPSExportTask

abisarvepalli:
- Remove empty export information from export
- Error handling when finding most intense peak


src/main/java/net/sf/mzmine/modules/peaklistmethods/io/gnpsexport/GNPSExportTask.java
```diff
@@ -53,7 +53,7 @@
 
   @Override
   public double getFinishedPercentage() {
-    return 0;
+    return 1;
```
tomas-pluskal: Could you add a proper progress estimate calculated from the number of peak list rows exported? 
See other modules for inspiration (it is a trivial calculation)

abisarvepalli:
Just updated the getFinishPercentage() method

PR #354, opened 2017-08-19
Fix SIRIUS export module

kaibioinfo:
- export all MS/MS scans instead of just the most intensive one
- add option to also export corresponding MS1 scans
- remove unused parameters
- cleaned up the corresponding class


tomas-pluskal:
Thanks, Kai!

PR #361, opened 2017-09-17
Fixed the error in ADAP3DTask when no peaks are found

asmirn1:
The algorithm still doesn't detect any peaks but at least it doesn't throw exceptions


tomas-pluskal:
I think the MSDK algorithm should never return null... This should be fixed at MSDK level, too.

asmirn1:
Yes, I'll fix it at msdk level too.

PR #362, opened 2017-09-18
New Spectral Deconvolution Module

asmirn1:
I've added a new spectral deconvolution module I've been working on. It's got much less parameters than the previous one. However, I kept the old version as well since the two versions are quite different and there might be people using the old one.


src/main/java/net/sf/mzmine/datamodel/MassSpectrum.java
```diff
@@ -38,7 +38,7 @@
     public Range<Double> getDataPointMZRange();
 
     /**
-     * Returns the top intensity data point. May return null if there are no
+     * Returns the top intensity data item. May return null if there are no
```
tomas-pluskal: Not sure what is the purpose of replacing all "point" with "item"?

src/main/java/net/sf/mzmine/desktop/preferences/MZminePreferences.java
```diff
@@ -56,7 +56,7 @@
 
     public static final FileNameParameter rExecPath = new FileNameParameter(
             "R executable path",
-            "Full R executable file path (If left blank, MZmine will try to find out automatically). On Windows, this should point to your R.exe file.");
+            "Full R executable file path (If left blank, MZmine will try to find out automatically). On Windows, this should item to your R.exe file.");
```
tomas-pluskal: "this should item to your R.exe file"? :)

tomas-pluskal:
Thanks for the new module!

The PR replaces all "point" with "item" across the whole project. I guess that was not the intention, or was it?

As for keeping the two modules, does it really make sense to have two ADAP deconvolution modules? If you think the new one performs better than the old one, it might be better to keep only the new one. If you think either of them may perform better under different circumstances, because the algorithms are different, then it might be better to give this one a different name?

asmirn1:
Sorry, replacing point to item was not my intension. I'll fix it.

Based on the tests I made, the two Spectral Deconvolution methods have similar performance. The advantage of the new method is the number of parameters. I would keep both methods at least for a while. For now, I can see three options:
1. Rename the two modules to "Spectral Deconvolution (Hierarchical Clustering)" and "Spectral Deconvolution (Blind Source Separation)"
2. Make a menu item "Spectral Deconvolution" with sub-menues "Hierarchical Clustering" and "Blind Source Separation"
3. Make a window "Spectral Deconvolution" with a parameter Algorithm (similar to "Chromatogram Deconvolution") where users can choose an algorithm.

I would prefer option 2 since it's easier to implement. What do you think?

tomas-pluskal:
Ok! Option 2 sounds good.

src/main/java/net/sf/mzmine/util/components/CombinedXICComponent.java
```diff
@@ -76,9 +76,11 @@ public CombinedXICComponent(Feature[] peaks, int id) {
             maxIntensity = Math.max(maxIntensity,
                     peak.getRawDataPointsIntensityRange().upperEndpoint());
             if (rtRange == null)
-                rtRange = peak.getDataFile().getDataRTRange();
+//                rtRange = peak.getDataFile().getDataRTRange();
+                rtRange = peak.getRawDataPointsRTRange();
```
tomas-pluskal: Please be careful about making changes to existing MZmine 2 modules as part of your pull requests. Such changes need to be carefully considered and discussed case-by-case. This change, for example, is unacceptable for me - it changes the existing plots into something that is much less useful (in my view). We can discuss what you are trying to achieve somewhere else, but in any case it does not belong into this PR.

src/main/java/net/sf/mzmine/modules/MZmineModuleCategory.java
```diff
@@ -30,6 +30,7 @@
     ISOTOPES("Isotopes"), //
     PEAKLIST("Peak list methods"), //
     PEAKLISTPICKING("Peak list processing"), //
+    SPECTRALDECONVOLUTION("Spectral Deconvolution"), //
```
tomas-pluskal: Spectral Deconvolution -> Spectral deconvolution

src/main/java/net/sf/mzmine/util/components/CombinedXICComponent.java
asmirn1: I'm sorry, we used the latter option in our lab, but you're right I should have discussed it with you before submitting these changes. Please, let me know where we can discuss it and I'll be glad to explain my reasoning.

tomas-pluskal: We can discuss by email or you can file an issue here or make a separate PR just for this change, whichever works best for you.

PR #374, opened 2017-10-24
Added RCaller support to main MZmine master branch.

golgauth:
Done with RCaller interface. 

* RsessionWrapper can handle both `Rsession` (**Rserve**) and `RCaller` (direct **Rscript** executable usage).
* RCaller has been made the default when initiating discussion with R (now, must explicitly ask for an Rsession session when instantiating the `RSessionWrapper`).
* We noticed quite a drawback on performances using RCaller, but should more stable.
* Next could be to remove Rserve stuffs (or not)... 

**_EDIT: OR, on the contrary, add a parameter to R featured stuffs, allowing the user to choose between the two engines (ability to fallback to RCaller when RServe fails, and keep Rserve when it works and higher performances are the priority)._**


tomas-pluskal:
Looks great. I like the idea of having RCaller or Rserve selection as a user parameter.

golgauth:
OK. I can provide to this. Would you prefer the user to choose on a per feature basis, or once and for all in MZmine preferences? If we stick to the proposal in the EDIT above, we better want to choose for each module that is concerned independently...

PR #375, opened 2017-10-26
Added RCaller/Rserve parameter on all features depending on R.

golgauth:


tomas-pluskal:
I was going to suggest that it might be easier to make this a Preferences parameter (similar to the R executable path that is currently there), but you were faster than me :) 
Thanks a lot for taking the effort!

golgauth:
Yep, if we consider the balance stability/performance depending on the usage of R made by every individual feature, I thought we might prefer being able to make a distinct choice per feature.
It was a bit longer to implement, that's right ;)

golgauth:
The Preferences parameter is still used. RCaller needs `Rscript` path which is deduced from `R` path (assuming they are in the same directory, no reason they wouldn't).

PR #378, opened 2017-11-07
Export merged MSMS and export single features

kaibioinfo:
- SiriusExportTask now provides three merging options:
 - no merge
 - merge only scans from the same sample
 - merge over all scans from all samples (all aligned columns within a feature row)
- merging is done by keeping the m/z of the most intensive peak but sum up intensities. Very close peaks from multiple spectra are merged together within 10 ppm

- add export function for single features. However, I have to check again what this strange fragment scan hashmap is doing I create at the beginning. Sounds not very efficiently if I do this for every single feature again.

### DIFF
--- a/src/main/java/net/sf/mzmine/modules/peaklistmethods/io/siriusexport/SiriusExportModule.java
+++ b/src/main/java/net/sf/mzmine/modules/peaklistmethods/io/siriusexport/SiriusExportModule.java
@@ -12,14 +12,18 @@
 
 package net.sf.mzmine.modules.peaklistmethods.io.siriusexport;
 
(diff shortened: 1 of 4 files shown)


tomas-pluskal:
Thanks!
When exporting a single feature, I think it would make more sense to export only the best MS and MS/MS scans, using
PeakListRow.getBestPeak().getRepresentativeScanNumber() and PeakListRow.getBestPeak().getMostIntenseFragmentScanNumber(). Not sure what the purpose of the HashMap is.

tomas-pluskal:
I'll merge this for now and do some testing.

PR #411, opened 2018-02-14
Added XYZDataset for EChartPanel

Annexhc:
Hi Tomas,
I ve found a bug in the modules Kendrick mass plot and Van Krevelen Diagram. The getSeriesCount() methods for the datasets had the wrong return value. Now the plots can be created very fast. 
I ve noticed the problem using the export data method from the drop down menu of the EChartPanel. The EChartPanel data export method only supported XYDatasets. I have added a small piece of code to handle XYZDatasets as well.
Best regards
Ansgar


tomas-pluskal:
Thanks, Ansgar!

PR #415, opened 2018-03-05
Added search bars to peaklist table

Akshaj5:


tomas-pluskal:
Thanks, looks pretty good to me!

src/main/java/net/sf/mzmine/modules/visualization/peaklisttable/PeakListTableWindow.java
```diff
@@ -105,6 +154,51 @@ public int getJScrollSizeHeight() {
...
+
+        TableRowSorter<PeakListTableModel> sorter = table.getTableRowSorter();
+
+        String textId = "(?i)" + filterTextId.getText();
+        String textMz = "(?i)" + filterTextMz.getText();
+        String textRt = "(?i)" + filterTextRt.getText();
+        String textIdentity = "(?i)" + filterTextIdentity.getText();
+        String textComment = "(?i)" + filterTextComment.getText();
+
+        rowSorters.add(RowFilter.regexFilter(textId,0));
```
tomas-pluskal: You could override the include() function of the returned RowFilter, and your overridden function can provide some debug output as to what is the value tested and what is the result of the test. That way you can debug the strange behavior of the filters.

tomas-pluskal:
Thanks, I will merge it and do some testing.

PR #421, opened 2018-03-21
Van Krevelen and Kendrick mass update

Annexhc:
Hi Tomas,
i have added an error message for the Van Krevelen module as you have suggested. Furthermore, I have updated the Kendrick mass module. The Kendrick mass defect base can now be entered as a sum formula, making the tool also interesting for other research fields apart from petrolomics and lipidomics. The code is now simpler and shorter as well. I have also updated the help.html accordingly. 
Best wishes,
Ansgar


tomas-pluskal:
Thanks a lot, Ansgar! 
I am traveling at the moment, but I plant to release a new MZmine version when I come back (end of next week).

PR #422, opened 2018-03-30
Master 2

golgauth:


golgauth:
- Uses hierarchical 'CACHED' clustering algorithm: [gnf.org](http://carrier.gnf.org/publications/cluster/) (allows huge elements sets clustering)
- Help for the module to follow soon
- Cleanup and more options to follow

TDI-tenderholt:
Is there a reason gnf clustering isn't available in Maven?

golgauth:
- Display of the dendrogram to be discussed: Do we want this feature, since the output visual won't be readable for large clustering trees?

golgauth:
> Is there a reason gnf clustering isn't available in Maven?

Well, I don't know but this is `Apache License - Version 2.0`.

src/main/java/net/sf/mzmine/modules/peaklistmethods/alignment/hierarchical/TriangularMatrix.java
```diff
@@ -0,0 +1,90 @@
...
+
+    public long sumFormula(long i) {
+        return (i * i + i) / 2;
+    }
+
+    public void print() {
+
+        for (int i = 0; i < getDimension(); i++) {
+
+            System.out.println("\n");
```
tomas-pluskal: Can we use Logger instead of the System.out.println() calls?

tomas-pluskal:
Thanks, Gauthier. What is the source of the gnf dependency? MZmine already uses Weka (https://www.cs.waikato.ac.nz/ml/weka/), which has implementations for various clustering algorithms. Is gnf something that Weka doesn't have?

src/main/java/net/sf/mzmine/modules/peaklistmethods/alignment/hierarchical/TriangularMatrix.java
golgauth: Sorry, I'll do a more advanced cleanup quickly. Many useless prints as well.

golgauth:
If you're OK, I'll find some time to cleanup all standard output stuffs tomorrow morning. Is that soon enough for you?

tomas-pluskal:
Tomorrow is totally fine. 

Regarding the clustering library, could you write some comments to the code where you would explain which libraries you tested and what was the final reasoning for the choice of gnf? Because that's valuable information, and if you don't document it, someone else will surely go through the same pain again in the future.

golgauth:
Sure ;)

I'll produce an abstract with all the libs. And may be add the WEKA version as an alternate clusterer option in the 'Hierarchical Alignment' module, such that anyone can make his own experience (If I remember well, WEKA provides more clustering strategies than gnf, such that it can be pertinent to add it, even if it is more limited in term of scalability).

PR #439, opened 2018-06-13
Sirius module

evgerher:
Development of GUI for new Sirius module for processing MS spectra.


pom.xml
```diff
@@ -618,6 +623,37 @@
...
+		<dependency>
+			<groupId>net.java.dev.jna</groupId>
+			<artifactId>jna</artifactId>
+			<version>4.5.0</version>
+		</dependency>
+
+		<dependency>
+			<groupId>org.openscience.cdk</groupId>
+			<artifactId>cdk-data</artifactId>
+			<version>2.0</version>
```
tomas-pluskal: Use ${cdk.version} here

pom.xml
```diff
@@ -618,6 +623,37 @@
...
+		<dependency>
+			<groupId>org.openscience.cdk</groupId>
+			<artifactId>cdk-interfaces</artifactId>
+			<version>${cdk.version}</version>
+		</dependency>
+
+		<dependency>
+			<groupId>org.openscience.cdk</groupId>
+			<artifactId>cdk-smiles</artifactId>
+			<version>2.0</version>
```
tomas-pluskal: Use ${cdk.version} here

src/main/java/net/sf/mzmine/modules/peaklistmethods/identification/sirius/IsotopeConstants.java
```diff
@@ -0,0 +1,65 @@
...
+import java.io.IOException;
+import org.openscience.cdk.config.IsotopeFactory;
+import org.openscience.cdk.config.Isotopes;
+import org.openscience.cdk.formula.MolecularFormulaRange;
+
+public class IsotopeConstants {
+  public static final int ISOTOPE_MAX = 100;
+  public static final int ISOTOPE_MIN = 0;
+
+  public static MolecularFormulaRange setDefaultCompounds() {
```
tomas-pluskal: This class looks a bit weird - it's called Constants, but it doesn't provide any useful constants. And the method setDefaultCompounds doesn't actually set anything, and has nothing to do with compounds. Think a bit more about the naming.

src/main/java/net/sf/mzmine/modules/peaklistmethods/identification/sirius/SingleRowIdentificationTask.java
```diff
@@ -0,0 +1,218 @@
...
+
+    ResultWindow window = new ResultWindow(peakListRow, searchedMass, this);
+    window.setTitle("Sirius makes fun " + massFormater.format(searchedMass) + " amu");
+    window.setVisible(true);
+
+    ConstraintsGenerator generator = new ConstraintsGenerator();
+    FormulaConstraints constraints = generator.generateConstraint(formulaRange);
+
+
+    double ppm = mzTolerance.getPpmTolerance();
```
tomas-pluskal: This needs updating. MZmine allows specifying both absolute (in m/z) and relative (in ppm) tolerances, and the final tolerance is always maximum of both values. But SIRIUS only works with ppm. So the correct approach is to obtain a mass range by calling mzTolerance.getToleranceRange(searchedMass) and then calculating the ppm value from that range.

src/main/java/net/sf/mzmine/modules/peaklistmethods/identification/sirius/AbstractParameters.java
```diff
@@ -0,0 +1,87 @@
...
+
+import java.io.IOException;
+import java.text.DecimalFormat;
+
+import net.sf.mzmine.parameters.parametertypes.IntegerParameter;
+import org.openscience.cdk.config.IsotopeFactory;
+import org.openscience.cdk.config.Isotopes;
+import org.openscience.cdk.formula.MolecularFormulaRange;
+
+public abstract class AbstractParameters extends SimpleParameterSet {
```
tomas-pluskal: AbstractParameters is a strange name for this class - how about SiriusParameters?

src/main/java/net/sf/mzmine/modules/peaklistmethods/identification/sirius/SingleRowIdentificationParameters.java
```diff
@@ -0,0 +1,53 @@
...
+import net.sf.mzmine.parameters.parametertypes.IntegerParameter;
+
+public class SingleRowIdentificationParameters extends SiriusParameters {
+  public static final IntegerParameter SIRIUS_CANDIDATES = new IntegerParameter(
+      "Number of candidates from Sirius method", "Maximum number of results to display", 5);
+
+  public static final IntegerParameter FINGERID_CANDIDATES = new IntegerParameter(
+      "Number of candidates from FingerId method", "Pass 0 to get all possible results", 5);
+
+  public static final DoubleParameter NEUTRAL_MASS = new DoubleParameter(
```
tomas-pluskal: This parameter should be called ION_MASS instead of NEUTRAL_MASS. And please use the m/z formatter for the parameter, see MZmineCore.getConfiguration().getMZFormat()

PR #444, opened 2018-06-19
Added MS Level to targeted peak filtering.

AndriusC:
Any changes or ideas on how the code can be improved will be appreciated.
(Re-push requesting as I messed around with branches and deleted the old one)

### DIFF
--- a/src/main/java/net/sf/mzmine/modules/rawdatamethods/peakpicking/targetedpeakdetection/TargetedPeakDetectionModule.java
+++ b/src/main/java/net/sf/mzmine/modules/rawdatamethods/peakpicking/targetedpeakdetection/TargetedPeakDetectionModule.java
@@ -1,17 +1,17 @@
 /*
  * Copyright 2006-2015 The MZmine 2 Development Team
- *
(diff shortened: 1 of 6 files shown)


tomas-pluskal:
Looks good, thanks! 

Just one concern: most of the diff are just formatting changes. We use Google Java Code Style (https://github.com/google/styleguide) to format our sources. Could you set your IDE to use that style and reformat the files that you modified in your pull requests?

AndriusC:
I am pretty certain that I installed a plugin that does exactly that thing, formats it in Google Java Code Style (https://plugins.jetbrains.com/plugin/8527-google-java-format) is this wrong? I never found that code style in IntelliJ by default.

tomas-pluskal:
Ah, sorry, I think you're right. The changes are probably because the previous version of the file was formatted with an old style.

AndriusC:
Although I'm not sure now, I got the code style from the link you gave me and it changed spacing(from the left side) and changed stuff like this:
if ... 
continue;

into

if .. {
continue;
}

PR #462, opened 2018-08-18
Isotope pattern recognition

SteffenHeu:
Hi Tomas,

I wrote a new Peak list method (IsotopePeakScanner) that can search a peak list for an isotope pattern of a given element combination and rate the results by comparing measured m/z shifts and intensity ratios to the calculates ones. The result will be assigned to the peaks as an isotope pattern. This improves the sum formula prediction's isotope pattern score if the compound contains more than 13C isotopes.

ips_sfp_results
Sum formula prediction of a compund called Pigment Red 112 (C24H16Cl3N3O2) with just 13C isotope recognition and now with Cl3 and 13C isotope recognition.

Furthermore, the peaks will be commented with a m/z shift (ppm) from the expected ones, an isotope combination resulting in the specific mass shift, intensity ratio and rating. The peak with the isotope combination with the lowest mass will be called _parent_ in the result peak list.

ips_result_example
Every isotope peak contains the ID of the parent, so you can sort them by comment and view all peaks of the isotope pattern at the same time. _Avg rating_ is a parameter that will calculate an averaged intensity of the given peaks in a specified mass list over all scans, but will only use those mass lists, that contain every recognized isotope peak at the same time.

_______

As a byproduct i created a new isotope pattern class based on the IsotopePattern interface in MZmine 2 that is able to calculate an isotope pattern on its own and will also yield a description of the data points since this was not possible before (ExtendedIsotopePattern.java). Another byproduct is NeutralLossFilter which works in a very similar way and scans a peak list for a specified neutral loss. At first this was integrated in IsotopePeakScanner, which is why the code is almost identical.

_______

The code was used to evaluate direct infusion and imaging data and tested with chromatographic data. I'd be happy for some comments or advice for improvements.

I've added help files to both modules. I hope this is as much information as you need!

The modules were created as a part of my bachelor thesis at University of Muenster, Germany. I was assisted by @robinschmid .

Best regards

Steffen Heuckeroth


tomas-pluskal:
Hi Steffen, 
Thanks for the contribution!
It seems there are some conflicts so this pull request cannot be merged at the moment. Could you please check it and resolve the conflicts against the current master branch?

SteffenHeu:
This caused github to mark a lot of files, but as unchanged. Is that a problem and should i create a new pull request with a new branch?

tomas-pluskal:
No need to create a new PR, this is fine.

src/main/java/net/sf/mzmine/modules/peaklistmethods/isotopes/isotopepeakscanner/Candidate.java
```diff
@@ -0,0 +1,283 @@
...
+ * 
+ * You should have received a copy of the GNU General Public License along with
+ * MZmine 2; if not, write to the Free Software Foundation, Inc., 51 Franklin St,
+ * Fifth Floor, Boston, MA 02110-1301 USA
+ */
+
+package net.sf.mzmine.modules.peaklistmethods.isotopes.isotopepeakscanner;
+
+import java.util.ArrayList;
+import org.jmol.util.Logger;
```
tomas-pluskal: You should not be using Logger from jmol. Please change to the JDK logger like in other MZmine modules.

src/main/java/net/sf/mzmine/modules/peaklistmethods/isotopes/isotopepeakscanner/Candidates.java
```diff
@@ -0,0 +1,460 @@
...
+import net.sf.mzmine.datamodel.MassList;
+import net.sf.mzmine.datamodel.PeakListRow;
+import net.sf.mzmine.datamodel.RawDataFile;
+import net.sf.mzmine.datamodel.Scan;
+import net.sf.mzmine.datamodel.impl.SimpleDataPoint;
+import net.sf.mzmine.datamodel.impl.SimpleIsotopePattern;
+import net.sf.mzmine.modules.peaklistmethods.isotopes.isotopepeakscanner.IsotopePeakScannerTask.RatingType;
+import net.sf.mzmine.parameters.parametertypes.tolerances.MZTolerance;
+
+public class Candidates {
```
tomas-pluskal: It would be great if you could add a brief JavaDoc description to each class, so we understand what the purpose of the class is.

src/main/java/net/sf/mzmine/modules/peaklistmethods/isotopes/isotopepeakscanner/IsotopePeakScannerParameters.java
```diff
@@ -0,0 +1,104 @@
...
+  public static final StringParameter element = new StringParameter("Element pattern",
+      "Element (combination) whose isotope pattern to be searched for. Please enter the two letter Symbol. (e.g. \"Gd\", \"Cl2Br\"S)");
+
+  public static final PercentParameter minAbundance = new PercentParameter("Minimum abundance",
+      "The minimum abundance (%) of Isotopes. Small values "
+          + "might increase accuracy but will decrease sensitivity.");
+
+  public static final DoubleParameter minHeight = new DoubleParameter("Minimum height",
+      "Minimum peak height to be considered as an isotope peak.",
+      NumberFormat.getNumberInstance(Locale.ENGLISH), 1E4);
```
tomas-pluskal: Use MZmineCore.getConfiguration().getIntensityFormat() instead

src/main/java/net/sf/mzmine/parameters/parametertypes/MassListParameter.java
```diff
@@ -30,6 +30,10 @@
 
     private String name, description, value;
 
+    public MassListParameter(String name1, String description1) {
+	this.name = name1;                  // i added this to make it possible to change the description of the mass list parameter
```
tomas-pluskal: You can remove the comments, or make them proper JavaDoc-style comments :)

tomas-pluskal:
Thanks!

PR #468, opened 2018-08-26
ChartGestures for JavaFX and CombinedXYPlots

robinschmid:
Hey Tomas,

as MZmine 3 will rely on JavaFX, I made the ChartGestures work with both swing and javaFX. The framework specific classes (only a few) are organised in the gui packages. Wrappers for MouseEvents and ChartViewer/ChartPanel connect swing and fx in a way that the same ChartGestureHandlers can be used for both. Another important addition, enables ChartGestures for CombinedXYPlots. 

The demos are important to quickly test new functions. I am also going to add this to MZmine 3 so that every contributer can use the newest version.

Still missing:
- CategoryPlot ChartGestures
- ZoomHistory for CombinedPlots

Best regards,
Robin


tomas-pluskal:
Hi Robin, 
Thank you, it looks great! Have you received any response from the JFreeChart people regarding integrating these functions directly into JFreeChart?

PR #472, opened 2018-09-02
Added error notifications via e-Mail

Annexhc:
Hi all,
I'm not sure if this is interesting for everyone, but I have implemented an e-mail notification system for errors. I have found that quite useful for long batch mode processing tasks. You can check a box in "Project->set preferences" called "send error e-mail notifications". You have to enter an email address, password, smtp server address, and smtp port. Ive tried it with an google mail account. It worked fine for me. The number of e-mails is restricted to 5 to avoid spam. Ive only implemented the smtp protocol because I dont have a lot of experience with automatic email transferring.
Does anyone thinks this is useful? I dont mind if its not going to be implemented in the master branch.
Best wishes,
Ansgar

### DIFF
--- a/src/main/java/net/sf/mzmine/desktop/impl/MainWindow.java
+++ b/src/main/java/net/sf/mzmine/desktop/impl/MainWindow.java
@@ -30,7 +30,6 @@ import java.io.File;
 import java.io.IOException;
 import java.util.logging.Level;
 import java.util.logging.Logger;
-
 import javax.annotation.Nonnull;
 import javax.help.HelpBroker;
 import javax.imageio.ImageIO;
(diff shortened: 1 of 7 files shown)


src/main/java/net/sf/mzmine/util/EMailUtil.java
```diff
@@ -0,0 +1,68 @@
...
+    Logger logger = Logger.getLogger("Mail Error");
+    try {
+      MimeMessage msg = new MimeMessage(session);
+      // set message headers
+      msg.addHeader("Content-type", "text/HTML; charset=UTF-8");
+      msg.addHeader("format", "flowed");
+      msg.addHeader("Content-Transfer-Encoding", "8bit");
+
+      msg.setFrom(new InternetAddress(toEmail, "MZmine"));
+      msg.setReplyTo(InternetAddress.parse("aiden.organa@gmail.com", false));
```
tomas-pluskal: Perhaps the email address should not be hard-coded here :)

src/main/java/net/sf/mzmine/desktop/impl/MainWindow.java
```diff
@@ -90,48 +92,76 @@ public HelpImpl getHelpImpl() {
...
   }
 
   public void displayMessage(Window window, String title, String msg, int type) {
 
+    // sending error message with a maximum of 5
+    if (MZminePreferences.sendErrorEMail.getValue() != null
+        && MZminePreferences.sendErrorEMail.getValue() && mailCounter < 5) {
+      ErrorMail errorMail = new ErrorMail();
+      try {
+        errorMail.sendErrorEmail(ErrorMailSettings.eMailAddress.getValue(),
```
tomas-pluskal: I think you should send the email in a separate thread, because the current implementation will block the GUI thread while the email is being sent. Perhaps the best is to make ErrorMail implement Runnable, then create a new Thread with it.

tomas-pluskal:
Thanks!

PR #477, opened 2018-09-17
ParameterSetupDialogWithEmptyPreview, new DataSet

SteffenHeu:
Hi Tomas,

this PR contains a new ParameterSetupDialog that can be used to easily implement any custom preview while the user is setting the parameters. In the beginning I did this only in a setup dialog for the Isotope peak scanner, but i thought it would be nicer to outsource this, so it can be used by programmers aswell.
I moved the JPanel pnlButtons inside ParameterSetupDialog, so it's possible to move the whole panel inside the ParameterSetupDialogWithEmptyPreview rather than every single button and creating a new panel.

Furthermore, this contains a DataSet that can be used to handle an ExtendedIsotopePattern. The dataset contains two XYSeries so data points below a given intensity can be drawn in a different colour, which allows the user to also see the isotope compositions that are not relevant yet, but might be with different parameters. This dataset will also contain the isotope compositions, so they can be used for a tool tip generator. This is also implemented in the SpectraToolTipGenerator already.

EDIT (additional info): The bars painted by the dataset will have a width (e.g. mergeWidth) according to the plot. (I thought this would be better for visualization of merged peaks than one-pixel-bars).

Implemented it can look like this:
grafik
(coming soon)

Best regards
Steffen


tomas-pluskal:
Looks good, thanks!

PR #479, opened 2018-09-17
Isotope peak scanner improvements

SteffenHeu:
Hey Tomas,

I improved the Isotope peak scanner. It still has all the functionality it had before, but it is now also able to automatically detect the amount of carbon atoms in an isotope pattern (parameter: Auto carbon). If the pattern contains heteroatoms that overlap with carbon peaks, the heteroatoms have to be entered, since they affect the relative intensities. The user enters a range of carbon atoms, then all isotope patterns will be calculated and the peak list will be compared to them. As long as the rating is ascending, the n+1 carbon pattern will be checked. If it starts to decrease, the last pattern will be selected as the best one. It is then added as a comment to the parent peak. This gives the user another hint while identifying unknown compounds.
I tested this will all the data i tested the previous IPS module. The results were exact sometimes, sometimes they were +-1 off the actual carbon number. The ions the new algorithm was tested with were in a m/z range of 320-690. I'm happy to test it with more data, if I get provided with some.
auto carbon example

The new parameter setup dialog (previous PR) is used to implement this preview, so the user sees all the peaks IPS will search for. The parameter Min intensity is displayed by the red, horizontal line and shows the threshold. Every peak with 0.2 times the minimum intensity is also calculated, so the user can see the effects of changig that parameter. Merge width is displayed by the thickness of the bars.
grafik

I hope you like the changes! As always I'm up for suggestions and improvements!

This PR is based on the previous one. If you want me to change something there, I'll do that first, of course.

Best regards
Steffen


tomas-pluskal:
Looks good. I will merge this and do some testing.

PR #480, opened 2018-09-17
Tool module class and Isotope pattern preview

SteffenHeu:
Hi Tomas,

I hope you and the others think this is useful. 
I added the module category "Tools" and a menu item. This also contains a Isotope pattern calculator that will plot a given isotope pattern. 
As you said, you can add them to a scan if you want, but i think sometimes its nice to see them seperately. Furthermore, the window will also contain a table with exact mass, intensity and isotope composition. That way the user can copy and export the data to excel or any other program as well.
grafik

This also depends on the dataset and setup dialog in one of the previous PRs. If you want any changes to be done let me know.

Best regards
Steffen


src/main/java/net/sf/mzmine/modules/tools/isotopepatternpreview/IsotopePatternPreviewParameters.java
```diff
@@ -0,0 +1,61 @@
...
+ * @author Steffen Heuckeroth steffen.heuckeroth@gmx.de / s_heuc03@uni-muenster.de
+ *
+ */
+public class IsotopePatternPreviewParameters extends SimpleParameterSet {
+
+  public static final StringParameter molecule = new StringParameter("Element/Molecule",
+      "The element/molecule to calculate the isotope pattern of. Enter a sum formula.");
+
+  public static final OptionalModuleParameter optionals =
+      new OptionalModuleParameter("Use custom paramters",
```
tomas-pluskal: paramters -> parameters

src/main/java/net/sf/mzmine/modules/tools/isotopepatternpreview/customparameters/IsotopePatternPreviewCustomParameters.java
```diff
@@ -0,0 +1,52 @@
...
+  
+  public static final DoubleParameter mergeWidth = new DoubleParameter("Merge width(Da)",
+      "This will be used to merge isotope compositions in the calculated isotope pattern if they overlap.",
+      MZmineCore.getConfiguration().getMZFormat(), 0.0005, 0.0d, 10.0d);
+  
+  public static final DoubleParameter minPatternIntensity =
+      new DoubleParameter("Min. pattern intensity",
+          "The minimum normalized intensity of a peak in the final calculated isotope pattern.\n"
+              + "Min = 0.0, Max = 0.99...",
+          new DecimalFormat("0.####"), 0.01, 0.0d, 0.99999);
```
tomas-pluskal: Why not use MZmineCore.getConfiguration().getIntensityFormat() ?

tomas-pluskal:
I am missing an option to specify the charge - you put m/z on the X axis but you assume z=1, which is not always the case.

SteffenHeu:
> 
> 
> I am missing an option to specify the charge - you put m/z on the X axis but you assume z=1, which is not always the case.

Ye I can do that. This was supposed to just display a molecule without charge, but i guess using charge would be more appropriate.

tomas-pluskal:
You need to consider the charge also for precise mass calculation (mass of electrons makes a difference).

src/main/java/net/sf/mzmine/modules/tools/isotopepatternpreview/customparameters/IsotopePatternPreviewCustomParameters.java
SteffenHeu: It will say 1.0E-2 after closing the dialog instead of 0.01. The y-axis of the plot is normalized, so the decimal format makes it more intuitive, I think. I will change it to MZmineCore.getConfiguration().getIntensityFormat() if you want, though.

tomas-pluskal: No that's fine, I didn't realize that this is relative intensity. Perhaps you should indicate that in the name ("Min. relative pattern intensity")

src/main/java/net/sf/mzmine/modules/tools/isotopepatternpreview/IsotopePatternPreviewDialog.java
```diff
@@ -0,0 +1,332 @@
...
+    if (!checkParameters())
+      return null;
+
+    if (molecule.equals(""))
+      return null;
+
+    logger.info("Calculating isotope pattern: " + molecule);
+
+    try {
+      pattern.setUpFromFormula(molecule, minAbundance, mergeWidth, minIntensity);
```
tomas-pluskal: I don't want to have two classes in MZmine that do exactly the same thing - why don't you reuse the existing calculateIsotopePattern() function in  net.sf.mzmine.modules.peaklistmethods.isotopes.isotopeprediction.IsotopePatternCalculator?

SteffenHeu: Because it does not yield a description of the isotope composition, which i needed for IPS and this. It keeps track of the isotope composition, while it's calculating the isotope pattern.
Since i changed a bit of the data structure (having a list instead of an array while calculating) I did not want to break other modules.

SteffenHeu:
Added. Chart and table will now say Exact mass / Da if charge = 0 and m/z any other time.

src/main/java/net/sf/mzmine/modules/tools/isotopepatternpreview/IsotopePatternPreviewDialog.java
SteffenHeu: Sounds reasonable. I will do that when im coming back from vacation (next wednesday), because i think it might take a few hours.

tomas-pluskal: In fact, it would be best to move that code to MSDK: 
https://github.com/msdk/msdk/blob/master/msdk-isotopes-isotopepattern/src/main/java/io/github/msdk/isotopes/isotopepattern/IsotopePatternGeneratorAlgorithm.java
Then we can remove the code from MZmine completely and just use the MSDK function.

tomas-pluskal: I am going to merge this PR now, and we can deal with these changes later.

SteffenHeu: I will look into it.

PR #492, opened 2018-10-14
Changes to the duplicate peak list row filter

robinschmid:
Hey Tomas,

this is a small change to the existing duplicate peaklist row filter. 
I kept the old mode (keep row with maximum average area) and added two more filter modes. 

One also filters the average RT and average m/z values to find duplicates and the other filters on a feature level. If two rows share one feature within RT and m/z tolerance in a raw data file, they are marked as duplicates.
From all duplicates, a consensus row is created.  All DETECTED features are favored over ESTIMATED and ESTIMATED over UNKNOWN. 

The detected features should be more comparable, as they were processed by multiple steps.

Best regards
Robin


src/main/java/net/sf/mzmine/modules/peaklistmethods/filtering/duplicatefilter/help/help.html
```diff
@@ -36,5 +51,11 @@ <h4>Method parameters</h4>
 
 </dl>
 
+
+        <p>
+        <h4>Aligned, gap-filled feature list prior and after duplicate row filter (new average).</h4>
+        <img src="prior.png" name="prior">
```
tomas-pluskal: The files you added are named prior.PNG and after.PNG - the capitalization does make a difference on unix platforms.

src/main/java/net/sf/mzmine/modules/peaklistmethods/filtering/duplicatefilter/DuplicateFilterTask.java
```diff
@@ -137,92 +139,153 @@ public void run() {
...
+                    break;
+                }
+              }
+            }
 
             // Duplicate peaks?
-            if (sameID && sameMZ && sameRT) {
-
-              peakListRows[secondRowIndex] = null;
+            if (sameID && sameMZRT) {
```
tomas-pluskal: It's hard to understand these nested loops and if statements. Any way to turn them into separate functions with self-explanatory names?

robinschmid:
Thanks for the code review. 
Changed the file names and moved some code into different methods to make it more readable.

src/main/java/net/sf/mzmine/modules/peaklistmethods/filtering/duplicatefilter/DuplicateFilterTask.java
```diff
@@ -137,92 +139,175 @@ public void run() {
...
+
+  /**
+   * Turns firstRow to consensus row. With all features with highest FeatureStatus:
+   * DETECTED>ESTIMATED>UNKNOWN Or the highest feature when comparing two ESTIMATED features
+   * 
+   * @param rawFiles
+   * @param firstRow
+   * @param secondRow
+   */
+  private void createConsensusRow(RawDataFile[] rawFiles, PeakListRow firstRow,
```
tomas-pluskal: I think this method should return a new PeakListRow, if you are creating a consensus of two rows. If you simply keep using the firstRow, then the firstRow might be referenced from both the old and the new peak lists, which would create data integrity issues, no?

src/main/java/net/sf/mzmine/modules/peaklistmethods/filtering/duplicatefilter/DuplicateFilterTask.java
```diff
@@ -137,92 +139,175 @@ public void run() {
...
+      PeakListRow secondRow) {
+    for (RawDataFile raw : rawFiles) {
+      Feature f2 = secondRow.getPeak(raw);
```
tomas-pluskal: The code looks a lot better when it is in a separate method, but still this block full of if() statements and brackets is a torture to read. 

I think you can make it more human-readable: 
- change `if (f2 != null) {` to `if (f2 == null) return;`, i.e. exit the method quickly rather than branching the code
- use a switch statement for `f2.getFeatureStatus()`, which will make the code flow more obvious

src/main/java/net/sf/mzmine/modules/peaklistmethods/filtering/duplicatefilter/DuplicateFilterTask.java
robinschmid: I have already copied the firstRow to keep the original peak list unchanged. This method only exchanges some of the features. As mentioned in the comments. The first row is always copied and inserted. 
You want me to change it?

src/main/java/net/sf/mzmine/modules/peaklistmethods/filtering/duplicatefilter/DuplicateFilterTask.java
```diff
@@ -167,8 +167,9 @@ private PeakList filterDuplicatePeakListRows(final PeakList origPeakList, final
     for (int firstRowIndex = 0; !isCanceled() && firstRowIndex < rowCount; firstRowIndex++) {
 
       final PeakListRow mainRow = peakListRows[firstRowIndex];
+
       if (mainRow != null) {
-        // need to copy to not alter the old peakList
+        // copy first row
         PeakListRow firstRow = copyRow(mainRow);
```
robinschmid: We need to copy each firstRow to be inserted later in the resulting peak list. 
As we can have multiple duplicate second rows, it is best to copy the row before the second loop. Then all features with a higher PeakStatus are exchanged.

src/main/java/net/sf/mzmine/modules/peaklistmethods/filtering/duplicatefilter/DuplicateFilterTask.java
```diff
@@ -229,25 +230,27 @@ private PeakList filterDuplicatePeakListRows(final PeakList origPeakList, final
...
+  private void createConsensusFirstRow(RawDataFile[] rawFiles, PeakListRow firstRow,
       PeakListRow secondRow) {
     for (RawDataFile raw : rawFiles) {
       Feature f2 = secondRow.getPeak(raw);
-      if (f2 != null) {
-        if (f2.getFeatureStatus().equals(FeatureStatus.DETECTED)) {
+      if (f2 == null)
+        continue;
+
+      switch (f2.getFeatureStatus()) {
```
robinschmid: I have changed it to `if(f2==null) continue;` 
Although I am not sure which style is more readable.  Still needs to loop through all raw data files. 

And I have added the switch block.

PR #493, opened 2018-10-14
Auto position of window

robinschmid:
Although MZmine is already trying to prevent this, my main window is sometimes off the screen at start. Therefore I have added a final check. It works in my cases. 

I had some problems after switching the number of screens from two to one. 

Cheers,
Robin

### DIFF
--- a/src/main/java/net/sf/mzmine/parameters/parametertypes/WindowSettingsParameter.java
+++ b/src/main/java/net/sf/mzmine/parameters/parametertypes/WindowSettingsParameter.java
@@ -162,7 +162,10 @@ public class WindowSettingsParameter implements Parameter<Object>, ComponentList
       frame.setExtendedState(Frame.MAXIMIZED_HORIZ | Frame.MAXIMIZED_VERT);
     }
 
+    // when still outside of screen
+    // e.g. changing from 2 screens to one
     if (!isOnScreen(frame)) {
+      // Maximise on screen 1
       frame.setLocation(0, 0);
       frame.setSize(1024, 800);
       frame.setExtendedState(Frame.MAXIMIZED_HORIZ | Frame.MAXIMIZED_VERT);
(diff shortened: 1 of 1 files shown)


src/main/java/net/sf/mzmine/parameters/parametertypes/WindowSettingsParameter.java
```diff
@@ -162,6 +161,32 @@ public void applySettingsToWindow(JFrame frame) {
...
+      // Maximise on screen 1
+      frame.setLocation(0, 0);
+      frame.setSize(1024, 800);
+      frame.setExtendedState(Frame.MAXIMIZED_HORIZ | Frame.MAXIMIZED_VERT);
+    }
+  }
+
+
+
+  public boolean isOnScreen(JFrame frame) {
```
tomas-pluskal: Make this private?

src/main/java/net/sf/mzmine/parameters/parametertypes/WindowSettingsParameter.java
```diff
@@ -162,6 +161,32 @@ public void applySettingsToWindow(JFrame frame) {
...
+    Rectangle virtualBounds = getVirtualBounds();
+    return virtualBounds.contains(frame.getBounds());
+  }
+
+  public static Rectangle getVirtualBounds() {
```
tomas-pluskal: Make this private?

src/main/java/net/sf/mzmine/parameters/parametertypes/WindowSettingsParameter.java
robinschmid: Could also become a util method and be moved to a WindowUtil class? Or public static

tomas-pluskal: Could be, but since there is no other use for it now, I think you can keep it here as a private method

robinschmid:
Changed to private

tomas-pluskal:
By the way, I noticed sometimes the parameter setup dialogs are shown behind the current window, but it only happens on Windows (I use a mac, so it does not affect me). Have you ever experienced that?

romchal:
yes, indeed, in Windows quite often the dialog windows tend to hide behind the current one

robinschmid:
Yeah same effect on windows. Still need to find an answer to this problem.

PR #494, opened 2018-10-15
Sort peak lists in join aligner

robinschmid:
Hey Tomas,

on Ming's suggestion, I have added an option to the join aligner to sort the peak lists prior to alignment. This should help in getting the same alignment results when reprocessing the data with the same parameters. 

The sort peak list module seems to be unfunctional in headless mode. Apart from that it might be a good idea to provide this option in the join aligner.

What do you think?

Cheers,
Robin

### DIFF
--- a/src/main/java/net/sf/mzmine/modules/peaklistmethods/alignment/join/JoinAlignerParameters.java
+++ b/src/main/java/net/sf/mzmine/modules/peaklistmethods/alignment/join/JoinAlignerParameters.java
@@ -33,6 +33,9 @@ public class JoinAlignerParameters extends SimpleParameterSet {
 
   public static final PeakListsParameter peakLists = new PeakListsParameter();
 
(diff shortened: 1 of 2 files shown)


tomas-pluskal:
Any idea why the "sort peak list module" is not working in headless mode? Wouldn't it make more sense to fix that module, instead of adding another parameter to join aligner? (how about the other aligner modules?)

robinschmid:
I have created a fix for the headless mode. Both sorting modules were accessing the TreeModels through the GUI and not the ProjectManager.

This small change to the join aligner is then not needed. Could still be handy though.

tomas-pluskal:
I would prefer to keep the sorting separate from the alignment, for two reasons:
- we have multiple different modules for alignment
- the sorting module can be further expanded in the future (e.g. sorting by sample groups etc.)

robinschmid:
Thats true

PR #495, opened 2018-10-16
Fix: Order raw files/pkl in headless mode

robinschmid:
Project was accessed through MainWindow not through ProjectManager


tomas-pluskal:
Awesome, thanks!

Issue #501, opened 2018-10-27
ADAP Chromatogram construction fails


nareshibt:
### Setup

I am reporting a problem with MZmine 2 version (2.34). 
My operating system is (Windows 7 64-bit).

I tried using both RAW file, converted it to MzXml format, the error appears on both instances.

[2:30:51|SEVERE|WorkerThread]: Unhandled exception java.lang.IllegalArgumentExce
ption: Invalid range: (305.1788330078125..305.1788330078125) while processing ta
sk Detecting chromatograms in D2_cells_alone.mzXML(java.lang.IllegalArgumentExce
ption: Invalid range: (305.1788330078125..305.1788330078125)@com.google.common.c
ollect.Range.<init>(Range.java:352))
java.lang.IllegalArgumentException: Invalid range: (305.1788330078125..305.17883
30078125)
        at com.google.common.collect.Range.<init>(Range.java:352)
        at com.google.common.collect.Range.create(Range.java:155)
        at com.google.common.collect.Range.open(Range.java:167)
        at net.sf.mzmine.modules.masslistmethods.ADAPchromatogrambuilder.ADAPChr
omatogramBuilderTask.run(ADAPChromatogramBuilderTask.java:312)
        at net.sf.mzmine.taskcontrol.impl.WorkerThread.run(WorkerThread.java:58)java -version

nareshibt:
I found a way around, I performed peak picking using msconvert (centroided the peaks). That solved the problem in this particular file (D2_cells_alone.mzXML).

But centroiding all my files is not helping. Because centroiding gives the same error on other files, which doesn't have the problem when used as mzxml file.

In this case I am using two different algorithms for mass detection based on the input file format

For RAW and mzxml files I will be using local maxima 
For centroided files (saved as mzxml file) I will be using centroided algorithm

But I am using ADAP chromatogram builder after mass detection on all my input files. 

Is this a right approach ?

asmirn1:
Hi @nareshibt could you share one of mzXML files you are using?

Your steps (mass detection and then chromatogram building) seems right to me. Let me try to process your files myself and see what causes the error.

nareshibt:
Hello @tomas-pluskal @asmirn1 and @duxiuxia thank you so much for taking your time to help me.

I have put two files in dropbox
https://www.dropbox.com/sh/91fdvl544olwfu1/AADnqyjvkYaBtO1rIv2EALLLa?dl=0

[...]

PR #502, opened 2018-10-29
Fix bug #501

asmirn1:
I've fixed bug #501. Now, file __D2_cells_alone.mzXml__ is processed normally without errors. I could not read file __blank.RAW__ (Getting error `java.lang.Exception: No scans found (NativeFileReadTask.java:144)`, but it's because of Mac + Wine + EXE combination)


tomas-pluskal:
Thanks!

PR #503, opened 2018-11-01
Implementation of CDK release into IsotopePatternPreview and IsotopePeakScanner

SteffenHeu:
Hi Tomas, 
I implemented the new CDK release and your suggestions. A formula check was put into FormulaUtils, i hope this covers everything.
I changed the IsotopePatternCalculator just a bit, so it uses the CDK merging by default, but I left the old merge-method of the calculator to not break other modules. This can be deleted in the future.
Also, I decided to leave SimpleIsotopePattern untouched and create a second IsotopePattern class, that can contain the isotope composition to not break other modules unexpectedly.

I wanted to make this two PRs, but implementing the changes needed for the preview would have broken IsotopePeakScanner, so i had to include it here.

Best regards
Steffen


tomas-pluskal:
Looks good, thanks!

PR #511, opened 2018-11-27
central m/z calculation after deconvolution

robinschmid:
Hey Tomas,

this PR adds a CenterMeasureParameter to the deconvolution. To avoid breakage of old batch files, this parameter is automatically set to the current MZmine standard (median).

However, this is a convenient way to set the function how mzmine calculates the central m/z value of all data points in a feature. 

- median is a good choice if you have many data points
- average or weighted average is good if ...
  - ... a feature has few data points
  - ... the mass accuracy drastically decreases for low abundant data points (LOG10 weighted-avg)

This PR also includes fix #510 

Cheers 
Robin

image
image

### DIFF
--- a/src/main/java/net/sf/mzmine/modules/peaklistmethods/peakpicking/deconvolution/ADAPpeakpicking/ADAPDetector.java
+++ b/src/main/java/net/sf/mzmine/modules/peaklistmethods/peakpicking/deconvolution/ADAPpeakpicking/ADAPDetector.java
@@ -21,26 +21,34 @@
 
 package net.sf.mzmine.modules.peaklistmethods.peakpicking.deconvolution.ADAPpeakpicking;
 
-// import static
-// net.sf.mzmine.modules.peaklistmethods.peakpicking.deconvolution.centwave.CentWaveDetectorParameters.INTEGRATION_METHOD;
-import static net.sf.mzmine.modules.peaklistmethods.peakpicking.deconvolution.ADAPpeakpicking.ADAPDetectorParameters.PEAK_DURATION;
-// import static
-// net.sf.mzmine.modules.peaklistmethods.peakpicking.deconvolution.centwave.CentWaveDetectorParameters.PEAK_SCALES;
-import static net.sf.mzmine.modules.peaklistmethods.peakpicking.deconvolution.ADAPpeakpicking.ADAPDetectorParameters.SN_THRESHOLD;
+import static dulab.adap.workflow.Deconvolution.DeconvoluteSignal;
+import static net.sf.mzmine.modules.peaklistmethods.peakpicking.deconvolution.ADAPpeakpicking.ADAPDetectorParameters.COEF_AREA_THRESHOLD;
 // import static
 // net.sf.mzmine.modules.peaklistmethods.peakpicking.deconvolution.ADAPpeakpicking.ADAPDetectorParameters.SHARP_THRESHOLD;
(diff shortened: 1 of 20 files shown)


tomas-pluskal:
Hi Robin,
I think this is a useful change! I am just not sure if giving people so many options for weighting is really useful - is there any practical reason for each of these? Over time I have realized that offering too much versatility can be detrimental, and it is often better to provide a few well-thought choices instead of many possible options.

robinschmid:
You are right. Same reason why I hate to go to a big grocery store.

I think median, mean and log10 weighted mean could be enough.
Sqrt might be nice as well. 
Linear is to much though.

I can change that tomorrow

robinschmid:
Hey Tomas,

I have changed it to these four options: MEDIAN, AVG and weighted AVG (LOG10, SQRT).

LOG10 is more noise level dependant, whereas SQRT shows the same behaviour for a 10-fold change. 

I think LOG10 is a good measure if the intensity range is large, as it does not overestimate the importance of the intensity on the mass accuracy.

LOG10 could be more robust if we divide the intensities by a noise level. This would make the noise level weight=1.
log10(value)-log10(noise)+1

To down regulate weighting for saturated signals with very high intensities, we could also set a maximum intensity as the weight 1



  |   |   |   |   |   | shifted to   noise level |   | Maximum capped |  
-- | -- | -- | -- | -- | -- | -- | -- | -- | --
Value | SQRT | LOG10 |   | Value | SQRT | LOG10 |   | Value | SQRT | LOG10
1000 | 31.6 | 3.0 |   | 10 | 3.2 | 1.0 |   | 10 | 0.141 | 0.371
2000 | 44.7 | 3.3 |   | 20 | 4.5 | 1.3 |   | 25 | 0.224 | 0.518
3000 | 54.8 | 3.5 |   | 30 | 5.5 | 1.5 |   | 50 | 0.316 | 0.629
4000 | 63.2 | 3.6 |   | 40 | 6.3 | 1.6 |   | 100 | 0.447 | 0.741
5000 | 70.7 | 3.7 |   | 50 | 7.1 | 1.7 |   | 250 | 0.707 | 0.888
6000 | 77.5 | 3.8 |   | 60 | 7.7 | 1.8 |   | 500 | 1.000 | 1.000
7000 | 83.7 | 3.8 |   | 70 | 8.4 | 1.8 |   | 750 | 1.000 | 1.000
8000 | 89.4 | 3.9 |   | 80 | 8.9 | 1.9 |   | 1000 | 1.000 | 1.000
9000 | 94.9 | 4.0 |   | 90 | 9.5 | 2.0 |   | 1250 | 1.000 | 1.000
10000 | 100.0 | 4.0 |   | 100 | 10.0 | 2.0 |   | 1500 | 1.000 | 1.000
50000 | 223.6 | 4.7 |   | 500 | 22.4 | 2.7 |   |   |   |  
  |   |   |   |   |   |   |   |   |   |  
  |   |   |   |   |   |   | Max intensity | 5.00E+02 | 22.4 | 2.7



image

image

tomas-pluskal:
My intuitive feeling is that one weighted average function should be enough. I agree that log10 is better, because it will give much less importance to noise-level signals.

I would go with 3 options: MEDIAN, AVG and LOG10-WEIGHTED-AVG.

robinschmid:
Thanks Tomas. 

This version is now limited to median, mean and auto (noise level corrected log10-weighted average)

The noise level is estimated as the minimum intensity in all chromatograms across all selected peaklists.

The weight in which a data point contributes to the central mz/z is then
Weight = log10(intensity) - log10(noise)

And it is capped at a maximum of 4

Issue #269, opened 2016-12-28
Formula prediction


tomas-pluskal:
The Identification>Formula prediction method assigns the formulas in the order as they are generated (which is quite unpredictable). It would be better to sort them first and then select the one with the lowest mass error or best isotope pattern match.

tomas-pluskal:
Hi Anja,
Use both the relative (ppm) and absolute (Da) tolerances. For example, if you set the window to "4ppm or 0.001 Da", the maximum of each value will always be used. That means for larger masses, the 4ppm will be used, while for small masses (where 4 ppm becomes too narrow), the 0.001 Da tolerance will be used. I think that will solve your issue.

anja-lauer:
It worked! Thank you so much :D

bruggemann-tropos:
Hello Tomas,

coming back to the initial idea of this thread:

> The Identification>Formula prediction method assigns the formulas in the order as they are generated (which is quite unpredictable). It would be better to sort them first and then select the one with the lowest mass error or best isotope pattern match.

Is the current version of MZmine still following this procedure? This would explain some strange assignments in my dataset.

Best wishes,
Martin

tomas-pluskal:
Hi Martin,
Yes, MZmine still follows the same procedure.,

tomas-pluskal:
Now fixed!

PR #515, opened 2018-11-29
Formula prediction

Annexhc:
Dear Tomas,
with this PR I address the issue #269. Now predicted formulas are sorted regarding their relative mass deviation prior to annotation.
best wishes,
Ansgar

### DIFF
--- a/src/main/java/net/sf/mzmine/modules/peaklistmethods/identification/formulapredictionpeaklist/FormulaPredictionPeakListTask.java
+++ b/src/main/java/net/sf/mzmine/modules/peaklistmethods/identification/formulapredictionpeaklist/FormulaPredictionPeakListTask.java
@@ -196,10 +196,11 @@ public class FormulaPredictionPeakListTask extends AbstractTask {
       }
       if (isCanceled())
         return;
-
       for (ResultFormula f : this.ResultingFormulas) {
-        SimplePeakIdentity newIdentity = new SimplePeakIdentity(f.getFormulaAsString());
-        row.addPeakIdentity(newIdentity, false);
+        if (f != null) {
+          SimplePeakIdentity newIdentity = new SimplePeakIdentity(f.getFormulaAsString());
+          row.addPeakIdentity(newIdentity, false);
+        }
       }
 
       finishedRows++;
--- a/src/main/java/net/sf/mzmine/modules/peaklistmethods/identification/formulapredictionpeaklist/FormulaPredictionPeakListTask.java
+++ b/src/main/java/net/sf/mzmine/modules/peaklistmethods/identification/formulapredictionpeaklist/FormulaPredictionPeakListTask.java
@@ -67,6 +67,7 @@ public class FormulaPredictionPeakListTask extends AbstractTask {
   private MZTolerance mzTolerance;
   private String message;
   private int totalRows, finishedRows;
+  private int maxBestFormulasPerPeak;
 
   /**
    *
@@ -108,6 +109,9 @@ public class FormulaPredictionPeakListTask extends AbstractTask {
         parameters.getParameter(FormulaPredictionPeakListParameters.elementalRatios).getValue();
     ratiosParameters = parameters.getParameter(FormulaPredictionPeakListParameters.elementalRatios)
         .getEmbeddedParameters();
+
+    maxBestFormulasPerPeak = parameters
+        .getParameter(FormulaPredictionPeakListParameters.maxBestFormulasPerPeak).getValue();
 
     message = "Formula Prediction";
   }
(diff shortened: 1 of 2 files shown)


tomas-pluskal:
Thanks, that is most welcome!

src/main/java/net/sf/mzmine/modules/peaklistmethods/identification/formulapredictionpeaklist/FormulaPredictionPeakListTask.java
```diff
@@ -163,18 +163,44 @@ public void run() {
           massRange.upperEndpoint(), elementCounts);
 
       IMolecularFormula cdkFormula;
+
+      // create a map to store ResultFormula and relative mass deviation for sorting
+      Map<Double, ResultFormula> possibleFormulas = new HashMap<Double, ResultFormula>();
```
tomas-pluskal: You can use a TreeMap here and keep all the results sorted. There is no need to create a HashMap and then convert it into a TreeMap later.

src/main/java/net/sf/mzmine/modules/peaklistmethods/identification/formulapredictionpeaklist/FormulaPredictionPeakListTask.java
```diff
@@ -163,18 +163,44 @@ public void run() {
...
       while ((cdkFormula = generator.getNextFormula()) != null) {
+        if (isCanceled())
+          return;
+
         // Mass is ok, so test other constraints
-        checkConstraints(cdkFormula, row);
+        ResultFormula resultEntry = checkConstraints(cdkFormula, row);
+
+        // calc rel mass deviation
+        Double relMassDev = ((searchedMass
```
tomas-pluskal: It is a good practice to avoid too many function calls in one line, because when an exception occurs, you won't know which function call had caused it. Why not separate these into three different lines
- MolecularFormulaManipulator.getString(cdkFormula))
- FormulaUtils.calculateExactMass(...)
- relMassDev = ...

src/main/java/net/sf/mzmine/modules/peaklistmethods/identification/formulapredictionpeaklist/FormulaPredictionPeakListTask.java
```diff
@@ -190,13 +216,13 @@ public void run() {
 
   }
 
-  private void checkConstraints(IMolecularFormula cdkFormula, PeakListRow peakListRow) {
+  private ResultFormula checkConstraints(IMolecularFormula cdkFormula, PeakListRow peakListRow) {
```
tomas-pluskal: Think it now makes more sense for this function to return boolean instead of ResultFormula

src/main/java/net/sf/mzmine/modules/peaklistmethods/identification/formulapredictionpeaklist/FormulaPredictionPeakListTask.java
Annexhc: agreed

src/main/java/net/sf/mzmine/modules/peaklistmethods/identification/formulapredictionpeaklist/FormulaPredictionPeakListTask.java
Annexhc: agreed

src/main/java/net/sf/mzmine/modules/peaklistmethods/identification/formulapredictionpeaklist/FormulaPredictionPeakListTask.java
Annexhc: hm I think it is the easiest way to get the formula of the checked cdkFormula. The old void method added the ResultFormula to a list. I have to loop through that list as well if I change the return type to boolean. Maybe we should just rename the method, e.g." getCheckedFormula"? What do you think?

Annexhc: okay, I have an other idea. Lets remove the "ResultingFormulas" List. It is not necessary, because only the formula is later used in the peaklist. All the other information the "ResultFormula" data type holds are not used.

tomas-pluskal: Yes, it seems the ResultingFormulas is redundant with your possibleFormulas.

tomas-pluskal: You can rename the checkConstraints() to getCheckedFormula()

tomas-pluskal:
Now that you added the sorting, it makes sense to restrict the formula generation to N best candidates. Could you add another parameter to the module, like "Max best formulas per peak"?

Annexhc:
sure!

robinschmid:
I also have something in the pipeline for this. However, it changes the structure how we assign formulae to a peak list row.

For sorting one can define weight factors for ppm distance, isotope score and ms/ms score. The weighted total score is then used for sorting.

robinschmid:
Important change though.

PR #519, opened 2018-12-03
IsotopePatternPreview improvements

SteffenHeu:
Hi Tomas,

the calculation of the isotope pattern was moved into an external thread. Furthermore I added an utility method to determine how big the isotope pattern will be. If it's above a certain size and there is a calculation running already, the new one will be queued. If it's a small isotope pattern, it will calculate it anyway. Also, if there is a queued formula already, the queue will be overwritten, to only calculate the latest pattern. Also, a status label was added.

To change the size of the charge text field, i had to add a method to IntegerComponent.

Best regards
Steffen


src/main/java/net/sf/mzmine/modules/tools/isotopepatternpreview/IsotopePatternPreviewTask.java
```diff
@@ -0,0 +1,115 @@
+package net.sf.mzmine.modules.tools.isotopepatternpreview;
```
tomas-pluskal: Please add the license header

src/main/java/net/sf/mzmine/modules/tools/isotopepatternpreview/IsotopePatternPreviewTask.java
```diff
@@ -0,0 +1,115 @@
...
+  
+  public void run() {
+    setStatus(TaskStatus.PROCESSING);
+    dialog.setStatus("Calculating...");
+    pattern = (ExtendedIsotopePattern) IsotopePatternCalculator.calculateIsotopePattern(formula, minIntensity, mergeWidth, charge, polarity, true);
+    if(pattern == null) {
+      logger.warning("Isotope pattern could not be calculated.");
+      return;
+    }
+    logger.info("Pattern " + pattern.getDescription() + " calculated.");
```
tomas-pluskal: Perhaps all these logging messages should go to logger.finest() instead of logger.info() (the INFO level messages are displayed in the status bar in the GUI, which I think is not necessary here)

tomas-pluskal:
Looks great, thanks!

Issue #527, opened 2018-12-14
export in GNPS


eligaro:
Hello,

My export for GNPS is not working properly. I am working with MzMine version 2.37 on a mac MacbookPro. 

I did select the Peak list filtered, then select in Peak List Method, Import/Export then Export for GNPS.
The process starts, there are three different tasks in the task window, but then nothing happen. The status of the processing bar say to zero and it doesn't move.

It did work last month though, and now not anymore.
Any idea why its not working? 

you can find my files:
https://www.dropbox.com/s/3iar965as27ct4w/A_Argyi_GE_EtOAC_FragDNP.mzmine?dl=0

Thanks a lot for your support,
Eliane

(*Please fill this in, and paste any relevant exceptions, log messages, and screenshots*)

tomas-pluskal:
@robinschmid Could you take a look at this issue?

tomas-pluskal:
@eligaro Please specify the exact parameter values you used to run the GNPS export module.

robinschmid:
Hey, 
I am going to try it. What are the exact parameters?
Did you use the new submit function?

robinschmid:
I think I figured out your problem. Go to Project/Set preferences 
Is your Number of Tasks set to 1? This would stop it as one task is waiting for the other ones to finish. 
Just set it to automatic. 

Tomas, I know that many people do not even know of this feature. Maybe we should show a message if this setting is way below the maximum.

![image](https://user-images.githubusercontent.com/10366914/50048220-1718f600-007b-11e9-9860-108df43d5f25.png)

eligaro:
Thanks a lot Robin.
That’s great. I indeed had this parameter set manually to 1. I modified it as suggested and export to GNPS and it seems to be partially working. The soft is still stuck on Task 1. I did export without submit to GNPS, because I want to do it offline first to make sure it works. When I click on OK, it quickly performs the task 2 and 3, but task 1 is still on processing and stay like that? Is this normal?
I then obtain two files: one name_quant.csv and another one name.mgf. what is the difference? Do I need to upload both into GNPS?
Thanks a lot
[cid:image001.png@01D495EB.4AFA0720]

robinschmid:
I have fixed this issue in PR #528 

Yeah there was a little bug that you needed to select at least one of the two options: open folder or submit to GNPS.

[...]

PR #528, opened 2018-12-17
Added getTaskPriority method for standard priority of a task

robinschmid:
This PR fixes issue #527 .
GNPS export/submit is not frozen on single core. 

Hey Tomas,

this is a small fix. I have added the method getTaskPriority to Task and AbstractTask will just return TaskPriority.NORMAL. Any other task can now implement this method to change the default priority. This specifies whether a task should restrain the max number of tasks or not. 

I think this change is more straight forward than adding a task with a specific task priority manually.
This:
`tasks.add(newTask);`
instead of:
`MZmineCore.getTaskController().addTask(newTask, TaskPriority.HIGH);`

I made sure that I dont change the Google statistics tracking.

Cheers,
Robin


tomas-pluskal:
Looks good, thanks!

PR #539, opened 2019-01-29
Multi MSMS dialog and mirror chart

robinschmid:
Hey Tomas,

a quick visual addition to mzmine and a small change to the SIRIUS export (from the peaklist popup menu). Now, all selected rows in a peaklist are exported to one mgf for SIRIUS, not only one. 

There are not too many changes. However, I have pushed all spectra visualisation modules into one package.

1. A multi MS/MS window for n MSMS scans
2. A mirror chart for comparison of 2 scans
(See images below)

image

image

Mirror chart of two MS/MS scans
image


tomas-pluskal:
Thanks, Robin!
It seems your PR has some conflicts with the most recent PR by @Annexhc, could you please resolve those conflicts?

tomas-pluskal:
Also, one inspiration: I think currently it is not very transparent how the SIRIUS (and also GNPS) exports work, I mean which spectra exactly get exported. If that's something you would be interested in improving, I imagine MZmine could have a ScanSelectionParameter (and corresponding ScanSelectionComponent), which would provide a checkbox-style list of scans that are included in the export. A great feature would be if the user could view and edit each spectrum before exporting (e.g., in a simple text format). Such parameter would be useful in many places in MZmine.

robinschmid:
I knew that there are some conflicts. Btw I really like ansgars changes. 
I am going to merge my changes into the latest master.

I am currently working on the sirius export. I might be able to develop a scanselection parameter. What do you mean by „to edit“

tomas-pluskal:
I was thinking that the user might want to modify the spectra manually in some cases. But now I realized the data is coming from the mass list, which itself is part of the parameter dialog, so editing capability would be way too complicated. Just adding a selection of scans is enough (e.g., in the form of MultiChoiceParameter). I think it would add more clarity to the export process.

src/main/java/net/sf/mzmine/chartbasics/gui/wrapper/GestureMouseAdapter.java
```diff
@@ -0,0 +1,47 @@
+package net.sf.mzmine.chartbasics.gui.wrapper;
```
tomas-pluskal: Please add license header

tomas-pluskal:
Looks great! I'll merge it and do some testing, too. 
I am planning to release a next version in the next few days, please let me know if you'd like to postpone for whatever reason.

src/main/java/net/sf/mzmine/chartbasics/gui/wrapper/GestureMouseAdapter.java
robinschmid: Done

tomas-pluskal:
Not sure what IDE you're using, but it is worth using the null checking capabilities (Eclipse can do that). It seems there are a few issues in the code:

screen shot 2019-02-01 at 12 15 50

src/main/java/net/sf/mzmine/modules/visualization/peaklisttable/PeakListTablePopupMenu.java
```diff
@@ -143,7 +143,7 @@ public PeakListTablePopupMenu(final PeakListTableWindow window, PeakListTable li
     showSpectrumItem = GUIUtils.addMenuItem(showMenu, "Mass spectrum", this);
     show2DItem = GUIUtils.addMenuItem(showMenu, "Peak in 2D", this);
     show3DItem = GUIUtils.addMenuItem(showMenu, "Peak in 3D", this);
-    showMSMSItem = GUIUtils.addMenuItem(showMenu, "Most intense MS/MS", this);
+    showMSMSItem = GUIUtils.addMenuItem(showMenu, "MS/MS (single/multiple)", this);
```
Annexhc: Great addition Robin! 
I am not 100% sure if I like the new label of this button. I think the information that the method selects the most intense MS/MS scan is more useful compared to the applicability on multiple rows. Maybe changing back to "Most intense" and adding "single/multiple row selection" to the tool tip? What do you think?

robinschmid:
> Not sure what IDE you're using, but it is worth using the null checking capabilities (Eclipse can do that). It seems there are a few issues in the code:
> 
> screen shot 2019-02-01 at 12 15 50

How did you get these error messages? All null cases are checked in this file - (with a separate method). I am using SonarLint for style and error checking and sometimes FindBugs.

Eclipse only shows me errors with the pom (I do not know how to get rid of those):
image

src/main/java/net/sf/mzmine/modules/visualization/peaklisttable/PeakListTablePopupMenu.java
tomas-pluskal: I agree

tomas-pluskal:
You can see these errors by going to the project properties in Eclipse -> Java Compiler -> Errors/Warnings -> Null analysis -> Enable annotation-based null analysis.
Then click on Configure to use the javax.annotation.Nonnull and Nullable annotations.
screen shot 2019-02-01 at 12 59 50

As for the pom.xml errors, try right click on project -> Maven -> Update Project and select Force update of Snapshots/Releases.

screen shot 2019-02-01 at 13 01 07

robinschmid:
Hey Tomas,
 
I have put some thinking into the Nullable annotation. They are definitely helpful to maintain and understand the code. However, in this case their analysis by eclipse gives wrong suggestions. The variables are declared as finals. Therefore, no other thread can change the value after initialisation of the object. Why is it still giving that error? 

  protected final @Nonnull IonModification adduct;
  protected final @Nullable IonModification mod;

There are many ways to handle NullPointerExceptions/null-objects. The question is, which one is the best for a broad community and an open source project? 

1. **final variables** (best option? but limited) 
2. **Optional<DataType>.of:** Good in combination with stream.flatmap (no need for null-checks in nested objects). Another thread could still change the Optional's content - so the problem is just shifted and the annotation neglected.
3. **Synchronized:** Bad, if there is no intent to use this object by parallel threads.

Any more options?


### Your original message:
FYI, there is a reason why this code produces a warning in Eclipse:

    if (this.mod != null)
      for (IonModification m : this.mod.getAdducts())
        allMod.add(m);

Since this.mod is @Nullable and not final, there is always a chance that it can be set to null value by another thread in between of the execution of the if (this.mod != null) statement and the this.mod.getAdducts() statement.

The proper workaround is something like:

    final IonModification modFinalCopy = this.mod;
    if (modFinalCopy != null)
      for (IonModification m : modFinalCopy.getAdducts())
        allMod.add(m);

tomas-pluskal:
You're right. If the fields are final, the warning is meaningless. In that case, let's just ignore it.

Final variables are generally good, so please use them whenever possible.

I don't think we need to use the Optional or synchronized workarounds here..

robinschmid:
It is weird though... Maybe I should ask in a forum. Or give this to the eclipse people.

Final is definitely the best case here and I am also trying to use it more often. 
The proposed workarounds were more of a general idea how to handle null-objects. 

Thanks for the discussion.

Cheers,
Robin

PR #542, opened 2019-02-04
Tomasmaster2

robinschmid:
See PR #539 

Changed title of show MS/MS and added tooltip.

### DIFF
--- a/src/main/java/net/sf/mzmine/modules/visualization/peaklisttable/PeakListTablePopupMenu.java
+++ b/src/main/java/net/sf/mzmine/modules/visualization/peaklisttable/PeakListTablePopupMenu.java
@@ -33,7 +33,6 @@ import javax.swing.JPopupMenu;
 import javax.swing.ListSelectionModel;
 import javax.swing.SwingUtilities;
 import javax.swing.table.AbstractTableModel;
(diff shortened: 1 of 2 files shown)


src/main/java/net/sf/mzmine/modules/visualization/peaklisttable/PeakListTablePopupMenu.java
```diff
@@ -33,7 +33,6 @@
 import javax.swing.ListSelectionModel;
 import javax.swing.SwingUtilities;
 import javax.swing.table.AbstractTableModel;
-import com.google.common.collect.Range;
```
tomas-pluskal: Removing this import causes compilation to fail. Why did you remove it?

robinschmid: Thats weird. I did not intentionally delete it. Maybe the auto import module in eclipse accidentally deleted it.
Sorry I thought I only did two small changes.
Can you change and reimport it?
I could do it tomorrow.

Thanks for your review. And sorry for that problem.

tomas-pluskal:
No problem, I'll fix that.

PR #545, opened 2019-02-22
DataPoint/Spectra processing

SteffenHeu:
Hey Tomas,

i added a framework/interface to directly process spectra/data points. In the future @Annexhc's SpectraIdentification modules can be implemented into this framework of course, to remove the duplicate code.

Basicly the creation of a new module will be similiar to every other MZmine module, it will just extend/implement different interfaces. You will still have a task, a module and a parameter set.
- Module: DataPointProcessingModule extends the MZmineModule interface. I had to create this, because the MZmineRunnableModule.runModule method needs a project which we dont have.
- Task: DataPointProcessingTask extends the AbstractTask class
- ParameterSet is just the same.

Every instance of SpectraPlot will have a DataPointProcessingController associated with it. The controller will pass the data to the first task. A StatusListener is added that automatically (and recursively) starts the next task when it has finished. The result of the previous tasks will be passed to the next one.
Results are stored in ProcessedDataPoint (extending SimpleDataPoint) which contains a list of DPPResult<T>. I already implemented basic result types, more can of course be added by implementing the interface DPPResult. The following tasks can of course access the results to improve their functionality.

Upon creation of the controller in SpectraPlot, it is passed to the DataPointProcessingManager, which will be a single instance. The manager keeps track of every controller created and starts/stops the execution of them.

The Setup of the parameters is done in the DPPSetupWindow. I used JavaFX for that, so it can be ported to MZmine 3 more easily. Every module can just be added once to the processing list.
setup window
(Btw i had to hardcode the initialisation of this window in MZmineCore and also avoid the destruction in [GUIUtils.closeAllWindows](https://github.com/SteffenHeu/mzmine2/blob/24651c089ecb75a180241e54dff10b07d906a158/src/main/java/net/sf/mzmine/util/GUIUtils.java#L95). Another desctruction issue is when the menu item Close all windows is clicked ([link](https://github.com/SteffenHeu/mzmine2/blob/24651c089ecb75a180241e54dff10b07d906a158/src/main/java/net/sf/mzmine/desktop/impl/WindowsMenu.java#L72)) I'd have to hardcode it here too, is there a reason this doesn't call the Guiutils method here?)
I'm not quite happy about this myself, but i couldn't think of a better way. I wanted to make the DPPSetupWindow a MZmineModule, but it does not have a parameter set class which leads to issues. )

The window creates a DataPointProcessingQueue similar to the BatchQueue, i just put all the saving and loading stuff in there.

A spectrum can look like this:
spectrum

Additionally I added two options to the preferences. A default processing queue can be loaded from an xml, and theres an option to enable/disable the processing.
preferences

I hope i did enough javadoc to make clear what everything does, if you need further insight just let me know! Of course I'm happy to do some changes to the code if you find any issues or think i can improve compatibility. 

Best regards

Steffen


tomas-pluskal:
Hi Steffen,
Thanks for the contribution, although I have to say I am a bit puzzled by this... what is the practical purpose of adding such global mechanism? 
Wouldn't it make more sense to add a button to the spectra visualizer like "Process spectrum" and give a selection of processing methods there?

SteffenHeu:
Hi Tomas,
the purpose is that you can quickly process a spectrum, without pressing any more buttons. The idea is, that more modules like database searches, better deisotoping, in-source-fragment searches or adduct ssearches get implemented into this base code, so you can quickly get good results by sum formula prediction and decide if is worth to look into a peak more specifically.
Also, currently all the SpectraIdentification modules run their own mass detection, and would have to run additional stuff like deisotoping by themselves, not passing the results to other modules that you might want to run on that spectrum aswell. So every module currently has to do the full work

This whole addition is entirely optional, so without enabling it, you wont notice it being there. 
I can of course add a button to the spectra visualizer and make it not start when loading the spectrum in any case, but that would take all the advantages of the automization away.

robinschmid:
It was my idea to have a mechanism that directly runs multiple analysis steps on spectra. MZmine has a high potential to have the best visualisation moduls. Many MZmine users and my collaborators still love to look into single spectra to verify and analyse the data. 


The next steps include:

- Add the whole processing steps: Thresholding, isotopes, adducts, formulas, lipid identification ...
- Create a new JavaFX window which holds spectra, chromatograms and everything else thats useful. 
    - Add smart controls and gestures to select...
    - ... spectra from chromatograms
    - ... chromatograms from an m/z range in a spectrum
    - ... MS/MS spectra from a spectrum or peak in chromatorgram

Cheers
Robin

Annexhc:
Hi Tomas, our idea behind this was the fully automated and fast analysis of spectra. As Robin pointed out most users still love to look into a spectrum. As Steffen mentioned, this could reduce overall processing resources. At the moment this is pretty fundamental regarding overall functionality, but it is the foundation to an extremely high level of automatization. By including more processing steps the user will directly get the ion identity of a signal with very high confidence, without doing anything by himself. I fully agree with you, that some buttons in the spectra toolbar would be useful, e.g. to toggle between annotations (m/z, sum formula...). Of course it is also debatable as you have mentioned where to put the button for the option (preferences or spectra toolbar). Another great example would be to add an online database search to the queue, because those are pretty slow.

tomas-pluskal:
Is this automatic processing function going to be used anywhere else then in the spectra visualizer?

SteffenHeu:
I'm not quite sure if I understand what exactly you mean. 
Right now a controller is implemented in every instance of SpectraPlot and automatically processes, if the option is enabled. The result is currently only displayed in the plot by a label generator.

Basically it can be imagined to do this for peak lists as well, but if someone wanted to use it on peak lists, they would use the batch mode, wouldn't they?

tomas-pluskal:
My question is whether you have any intention for this controller to be used in other modules than the spectra visualizer.

The reason why I am not so happy about this change is because it goes against the philosophy of modularity. But my impression so far is that all these data processing functions are used only in the spectra visualizer, and therefore I think you should move all the functions to the spectra visualizer sub-package and activate the functions from the spectra visualizer window, not from the global Preferences dialog.

SteffenHeu:
I can do that, yes.
I tried to make everything as modular as possible, but if you have further suggestions just let me know ;).

robinschmid:
The idea was to make it as general as possible. To allow the processing of a DataPoint array which can originate from any source. This can be in a visualisation module or in any processing method. This was the initial intend and is not relying on the controller.

The main idea of the controller is to connect the logic with a visualisation module. The manager class is organising the task processing to only allow a specific number of parallel tasks. (Not to flood MZmine with spectra identification tasks, if the user opens a hundred spectra) 

The SpectraPlot is great and it should be expanded. However, maybe we are going to have more than one visualisation module of spectra, or more generally speaking, of datapoints. 

- MS/MS visualiser
- Mirror spectra
- Multi MS/MS window

For example, if we create a window with a chromatogram and an optional number of spectra, this window would need to create the controller and stop them if a spectrum has changed.

tomas-pluskal:
@SteffenHeu Let's start with adding your functions to the spectra visualizer package. How about adding a checkbox to the bottom toolbar like "[ ] Enable processing (...)" and setup the processing parameters after the user clicks the (...) button.

@robinschmid Actually, this is what I am trying to achieve with MSDK - all the algorithms will be in MSDK, and MZmine and other software tools can use them for various things.

As for the spectra visualizer, I don't think we should have multiple modules that do the same thing (i.e., plot m/z vs intensity). A better strategy is to make a "modular" spectra visualizer, which can plot multiple different data types as layers. In fact I already did some work in that direction a few years ago, which is in the mzmine3 tree, but unfortunately the mzmine3 tree is not functional at this point. Perhaps I will backport that module to mzmine2 and then we can take a look at it together.

robinschmid:
We should definitely port this to MSDK. However, MZmine 2 is currently using the MZmine modules and this could be a stretch to program it into the MSDK and MZmine 2 right now. 

If I get this right, your main point is to move the parameter setup and enabling to the spectra plot, correct? The automatic nature and task management is kind of fine?

tomas-pluskal:
Yes, it is fine. I think at this point the GUI as well as the source code (packages) should be integrated into the spectra visualizer, unless you guys have a clear use case in another module.

robinschmid:
Yeah sounds good to me. @SteffenHeu ?

tomas-pluskal:
One more question: what is the reason you don’t use the existing task manager for the job management?

src/main/java/net/sf/mzmine/modules/datapointprocessing/DataPointProcessingController.java
```diff
@@ -0,0 +1,367 @@
...
+                case CANCELED:
+                  setStatus(ControllerStatus.CANCELED);
+                  break;
+              }
+            }
+          });
+
+      setCurrentTask((DataPointProcessingTask) t); // maybe we need this some time
+      logger.info("Start processing of " + t.getClass().getName());
+      MZmineCore.getTaskController().addTask(t);
```
robinschmid: I think it is using the task controller.
The idea was to prevent any spamming of hundreds of tasks to the task controller

SteffenHeu: Yes, the Manager just "manages" the controllers. It adds them to the task controller when there is no more than (currently) 3 data point tasks running at the moment.

SteffenHeu:
> 
> 
> Yeah sounds good to me. @SteffenHeu ?

I will do that tomorrow. Maybe i'll have to do some changes but i don't know yet. It shouldn't be too much of a stretch.

tomas-pluskal:
@SteffenHeu Could you explain briefly in a few sentences how the whole controller works?

SteffenHeu:
Yeah:
So, a controller is created in the addDataset method of spectra plot, if the processing option is enabled. Each plot has one controller. The controller controls the task execution, _after_ it has been started by the execute() method, it is also used to cancel the execution of tasks, if the plot changes its dataset. 
When the execute() method is called, the first task of the processing queue is started and a status listener is added. This status listener starts the execution of the next task in the queue, when the first one has reached TaskStatus.FINISHED. If there are no more task in the queue it will display the results of the last task in the associated plot.

One has to note, that the execute method is _not_ called by the spectra plot. The spectra plot just adds the controller to the DataPointProcessingManager's waiting list. If there are less than three controllers running, the execution of the controller is started automatically, by adding the task to the regular TaskController. If not, it will be started, when a running controller-task has finished. This was done to not flood the TaskController with too many tasks, if a window with multiple plots is opened.

SteffenHeu:
Furthermore, i was just thinking that it might be better, to add a abstract method displayResults() to the DataPointProcessing class that gets called when a task finishes. An additional parameter would then specify if the results of that method shall be displayed or not. This might make it easier to implement visualization of more complex results, because the controller would just call the method of the task if specified. Else a modification of the controller might be needed every time a new result type is added.

tomas-pluskal:
@SteffenHeu Thanks a lot for the explanation!

I want to say that I like the idea of this kind of quick pre-processing of the spectra directly from the spectra visualizer, MZmine users will surely appreciate it. But I am not so convinced about the implementation you have chosen, from software design perspective. It seems the controllers you added are redundant, performing a very similar function as the task controller. And I am not even sure what happens to your listeners when things start breaking down. For example: what if one of the tasks in your queue fails? Or what if the user starts switching scans quickly (e.g., using the right arrow button) - will each newly loaded scan create a long queue of tasks, and will those tasks get cancelled automatically when next scan is loaded? If the spectra processing is activated, will it affect preview plots (e.g., in mass detection module)?

I have not been able to get your functions to work, because the whole MZmine freezes when I click the "Set parameters" button on any of the processing tasks. Probably some threading clash between the Swing and JavaFX control threads?

A few points for discussion:

1. Does the pre-processing really need to be automatic? What if you add a simple button to the bottom of the spectra visualizer, like "Process this spectrum"? When the user clicks this button, it would start a **single** processing meta-task (similar to the batch mode meta-task), which would then execute its individual sub-tasks as configured. This way everything could be handled by the existing task manager and the user would always see what is running and have control over it. Also, most importantly, tasks would not be created in the background without the users's specific action. On the other hand, activating the processing would be a single click, so it would be still very straightforward and quick.

2. I see you added an arbitrary selection of two mass detection methods, but not others - I think this should be designed in a better way to reuse the existing dialogs and parameters. Again, take a look at the batch mode - it reuses all the ParameterSets of the individual modules, so it does not need to create new ways to define parameters. It would be much better to use the existing ParameterSetupDialog for configuring your functions, rather than creating a completely new GUI window for it (of course you can create a new component, but again it should be done in a consistent way with other modules, by creating a new Parameter and its editing component).

3. As I said above, I have not been able to actually run the processing, so I haven't seen how it behaves in action. But thinking a bit more about the "layered" design of the spectra visualizer - does your module add a new dataset/layer to the plot with the processed spectrum? What if multiple processing steps are activated, does it add multiple new datasets?

SteffenHeu:
I just have to say, the three of you guys are most likely way better software engineers, so i'll change the code if you want anything changed. I just want to explain why I did it this way, first:

> It seems the controllers you added are redundant, performing a very similar function as the task controller.

It does perform a similiar function, yes. But you would still need the functionality of the controller, e.g. waiting for the first task to finish, before you start the next one. I just thought the implementation with a listener in the controller, to be notified if a task changes it's status would be the best. Else, you would have to check every x ms and unnecessarily delay the processing.

> And I am not even sure what happens to your listeners when things start breaking down. For example: what if one of the tasks in your queue fails?

If a task status changes to error/is canceled, the next task is not started and the execution of the whole queue of that plot is canceled. 

> Or what if the user starts switching scans quickly (e.g., using the right arrow button) - will each newly loaded scan create a long queue of tasks, and will those tasks get cancelled automatically when next scan is loaded?

Tasks are created in that scenario, yes. But the method removeAllDataSets, which is called when a new scan is loaded cancels the tasks, there should not be a problem, if the task itself supports cancelling during the run method.

>If the spectra processing is activated, will it affect preview plots (e.g., in mass detection module)?
Good point, i will have to change that. Currently it is, yes.

> I have not been able to get your functions to work, because the whole MZmine freezes when I click the "Set parameters" button on any of the processing tasks. Probably some threading clash between the Swing and JavaFX control threads?

I could not reproduce that error, do you get any exception or logger message?

> 1. Does the pre-processing really need to be automatic? What if you add a simple button to the bottom of the spectra visualizer, like "Process this spectrum"? When the user clicks this button, it would start a **single** processing meta-task (similar to the batch mode meta-task), which would then execute its individual sub-tasks as configured. This way everything could be handled by the existing task manager and the user would always see what is running and have control over it. Also, most importantly, tasks would not be created in the background without the users's specific action. On the other hand, activating the processing would be a single click, so it would be still very straightforward and quick.

I feel like having to press a button everytime will take away some of the advantages of the automization. Right now you could just quickly skim through the spectras, while the annotation is done automatically. With an extra button to click it will take something of the comfort away. 
But that meta task would basically be the same as the current controller, or? With the difference the modification of the plot comes from another thread.

> 2. I see you added an arbitrary selection of two mass detection methods, but not others - I think this should be designed in a better way to reuse the existing dialogs and parameters. Again, take a look at the batch mode - it reuses all the ParameterSets of the individual modules, so it does not need to create new ways to define parameters. It would be much better to use the existing ParameterSetupDialog for configuring your functions, rather than creating a completely new GUI window for it (of course you can create a new component, but again it should be done in a consistent way with other modules, by creating a new Parameter and its editing component).

I agree that adding more mass selectors would be beneficial.
The problem with the parameter setup dialogs is, that most of the parameter sets have parameters like RT tolerance, which won't be used in the scan processing modules and might be misleading. On the other hand, some might lack a parameter like an addition noise level, if you just wanted to do a formula prediction for peaks above a given intensity. That's why i thought keeping the general architecture of the mzmine modules consisting of a module, a parameter set and a task might be better than eagerly trying to detour the way through by removing that parameter from the setup dialog.

> 3. As I said above, I have not been able to actually run the processing, so I haven't seen how it behaves in action. But thinking a bit more about the "layered" design of the spectra visualizer - does your module add a new dataset/layer to the plot with the processed spectrum? What if multiple processing steps are activated, does it add multiple new datasets?

I really don't know why it's not running for you, we tried it yesterday on robins computer and it worked. I'll have to look into that. Can you rightclick in the tree view and set the parameters? Does everything else work, like adding/removing new modules to/from the list?
Since the latest commit, the DataPointProcessingTask has an abstract method displayResults() through which the results will be displayed. I added a method to the controller to check, if the current task is the last one, in that case the result should be displayed in any way. In other cases i added a parameter to the existing modules in which the user can check, if they want to add a new Dataset with the results of each step.

tomas-pluskal:
I tried using the right click in the tree view, and it immediately froze. This is the last message in the console:
```
[17:26:22|FINEST|SpectraVisualizerWindow]: Loading scan #204 from 4C_2H_05.raw for spectra visualizer
[17:26:22|FINEST|SpectraBottomPanel]: Rebuilding the peak list selector
[17:26:22|FINEST|SpectraVisualizerWindow]: Loading a peak list 4C_2H_05.raw chromatograms to a spectrum window Spectrum: [4C_2H_05.raw; scan #204]
[17:26:42|FINEST|DPPSetupWindowController]: Button[id=btnAdd, styleClass=button]'Add' clicked.
[17:26:42|FINEST|DPPSetupWindowController]: Added module Exact mass/Centroid mass detection to processing list.
[17:26:45|FINEST|DPPSetupWindowController]: Double clicked item in processing tree view.
```

Adding and removing modules works fine.

SteffenHeu:
I removed the java fx functionality and added a new component and parameter. Please tell me if it works for you now.

robinschmid:
Sorry for my absence. I was on a conference and totally tied up with work. 
@tomas-pluskal you're testing it on a mac right? The javafx GUI worked for me, maybe it is something platform or JRE related? 
However, the new swing GUI works for me.

I have read through some of the code. The task management seems to work fine when skipping through the spectra at full speed - only one thread. Moreover, the processing and annotation works like a charme, without crashing or slowing down MZmine. 

The parameter setup in now better placed inside the spectraplot. I personally prefer the checkbox to activate automatic processing rather than pressing a button.

**About the mass detection:**
Don't you have everything in place to add the missing mass detectors? You are setting the noise level as a new parameter to then give it the mass detector paramters. You can delete this parameter and just directly use the mass detector parameters and allow any mass detector (see the general mass detector parameters).

**Module structure**
It would be nice reduce the need to implement a special Module, Parameterset and Task for this type of processing modules. However, as many modules (e.g., the deisotoper) are implemented to work on a peaklist and not a DataPoint array, we definitely need some new structures. Maybe it is better to transform your DataPointProcessingModule into a module interface which can create a task for a peaklist or DataPoint array. Your DataPointProcessingTask could then be a task that can handle both inputs.

This structure would enforce people to think about their way of implementing a new processing method.

Cheers
Robin

tomas-pluskal:
Hi @SteffenHeu,
It looks much better now, and I can finally test how it works :). A few comments:
- There are two different buttons for enabling the processing, and they're not synchronized. Please use only one button, and perhaps move it from the toolbar to the bottom panel next to the peak list selector:
screen shot 2019-03-05 at 16 34 26
- The meaning of "Default spectra processing queue" is totally unclear to me
- As Robin suggested, it would be better if you reuse the existing parameters (e.g., MassDetectionParameters.massDetector) instead of designing new ones
- It is a bit unclear to me if the processing modules are executed sequentially (i.e., first mass detection, then formula prediction executed on the mass list) or independently
- The parameter setup dialog has no Help, but that can be added once the functions are finalized
- This happened when I tried using the formula prediction:
````[16:32:27|SEVERE|WorkerThread]: Unhandled exception java.lang.NullPointerException while processing task Predicts sum formulas for Scan #424(java.lang.NullPointerException@net.sf.mzmine.modules.visualization.spectra.simplespectra.datapointprocessing.sumformulaprediction.DPPSumFormulaPredictionTask.run(DPPSumFormulaPredictionTask.java:171))
java.lang.NullPointerException
	at net.sf.mzmine.modules.visualization.spectra.simplespectra.datapointprocessing.sumformulaprediction.DPPSumFormulaPredictionTask.run(DPPSumFormulaPredictionTask.java:171)
	at net.sf.mzmine.taskcontrol.impl.WorkerThread.run(WorkerThread.java:58)
````

**Comment by tomas-pluskal** [comment]
@robinschmid Yes, I am using a mac. I'm sure it is some threading issue between JavaFX and Swing, but there should be a way for both to coexist properly.

I agree that the module should be structured in a way that it doesn't generate redundant functionality. There is the option to modify the existing modules (e.g, the isotope grouper) to process both spectra and peak lists, and then just refer to that module in this visualizer.

**Comment by SteffenHeu** [comment]
Hey Tomas,
1. You are right. I will have to remove the checkbox from the SpectraVisualizerWindow, otherwise I wont be able to use it in the ParameterSet. I will move the button to the bottom panel.
2. It's a leftover from when the queue wasn't a regular parameter but managed in the javafx window. I just quickly ported the functionality yesterday. Removing it is on my to-do list.
3. For the mass detectors yes, it does make sense there. The problem with other parameter sets is, as i mentioned before: Some contain parameters like peak lists, raw data files, mass lists, retention time tolerances and so on. These parameter types have no use in this application and will therefore irritate the user. Reusing existing parameter sets should be possible already, since the parameter set is assigned to the module via the getParameterSetClass method which the DataPointProcessingModule has to implement by extending the MZmineModule interface. The existing parameter set could be passed here. 
But this will lead to a lacking of the parameters used to decide whether or not the results of that module should be displayed and which color the dataset should have.
A possible workaround **could** be, that a module makes a **new** parameter set containing just the color and displayResults parameter _and then_ integrates the existing paramaters as embedded parameter set, which will lead to two dialogs though. Also this won't solve the issue with useless parameters for this application.

Another important point: We might need special parameters just for this application. For example, if we run an adduct search. Let's say we detect M+H+, M+NH4+ and M+Na+ of a molecule. The adduct result is then stored into each ProcessedDataPoint. If we now run a sum formula prediction, we obviously want to use this adduct result as ionization type for the prediction, which would make the ionization parameter currently used in the prediction unnecessary, or we would need the possibility to select "dynamic/base on results", which would only be applicable in this case.

I hope you understand why I'm a bit hesitant about reusing existing parameter sets.

4. They are executed sequentially from top to bottom. The next module will work on the results of the previous one.

6. Can you tell me which parameters you used? I couldn't reproduce that error yet.

Regarding the module structure:
Existing modules can implement the DataPointProcessing interface next to the MZmineRunnable module, which would avoid having two modules. (Eclipse does not grumble about both extending MZmineModule. Could this be an issue?)
I think the tasks doing the processing should not be put into a single file. Some run-methods are pretty long already, having another branching between data points and peak lists would reduce the clarity even more. This would either need a complete reconsideration of how the tasks are designed e.g. putting a lot of the maths into utility classes to reduce the size of a single file. Although this could be beneficial, i think it's not worth the effort for MZmine 2 but rather for MZmine 3?

Please let me know, what you guys think!

**Comment by tomas-pluskal** [comment]
Hi Steffen,
I suppose when I talked about reusing parameters, I meant the actual Parameter instances, not the whole ParameterSets, for the same reasons you describe.

I think it would make most sense to add all your parameters into the SpectraVisualizerParameters as an OptionalModuleParameter. 

The help for this function should clearly explain that the processing steps are executed sequentially.

Regarding the exception, you can reproduce it this way:
- load the sample dataset (sample-2.3.mzmine)
- open scan #1 in BLANK sample
- add 13-C Isotope grouper to the Processing Steps with the following parameters:

Screen Shot 2019-03-09 at 15 23 26

I get the following exception with your latest code:
```
Exception in thread "AWT-EventQueue-0" java.lang.NullPointerException
	at net.sf.mzmine.modules.visualization.spectra.simplespectra.datapointprocessing.DataPointProcessingManager.clearProcessingSteps(DataPointProcessingManager.java:297)
	at net.sf.mzmine.parameters.parametertypes.ProcessingComponent.sendQueue(ProcessingComponent.java:299)
	at net.sf.mzmine.parameters.parametertypes.ProcessingComponent.setParameters(ProcessingComponent.java:196)
	at net.sf.mzmine.parameters.parametertypes.ProcessingComponent.actionPerformed(ProcessingComponent.java:96)
	at javax.swing.AbstractButton.fireActionPerformed(AbstractButton.java:2022)
	at javax.swing.AbstractButton$Handler.actionPerformed(AbstractButton.java:2348)
	at javax.swing.DefaultButtonModel.fireActionPerformed(DefaultButtonModel.java:402)
	at javax.swing.DefaultButtonModel.setPressed(DefaultButtonModel.java:259)
	at javax.swing.plaf.basic.BasicButtonListener.mouseReleased(BasicButtonListener.java:252)
	at java.awt.Component.processMouseEvent(Component.java:6539)
	at javax.swing.JComponent.processMouseEvent(JComponent.java:3324)
	at java.awt.Component.processEvent(Component.java:6304)
	at java.awt.Container.processEvent(Container.java:2239)
	at java.awt.Component.dispatchEventImpl(Component.java:4889)
	at java.awt.Container.dispatchEventImpl(Container.java:2297)
	at java.awt.Component.dispatchEvent(Component.java:4711)
	at java.awt.LightweightDispatcher.retargetMouseEvent(Container.java:4904)
	at java.awt.LightweightDispatcher.processMouseEvent(Container.java:4535)
	at java.awt.LightweightDispatcher.dispatchEvent(Container.java:4476)
	at java.awt.Container.dispatchEventImpl(Container.java:2283)
	at java.awt.Window.dispatchEventImpl(Window.java:2746)
	at java.awt.Component.dispatchEvent(Component.java:4711)
	at java.awt.EventQueue.dispatchEventImpl(EventQueue.java:760)
	at java.awt.EventQueue.access$500(EventQueue.java:97)
	at java.awt.EventQueue$3.run(EventQueue.java:709)
	at java.awt.EventQueue$3.run(EventQueue.java:703)
	at java.security.AccessController.doPrivileged(Native Method)
	at java.security.ProtectionDomain$JavaSecurityAccessImpl.doIntersectionPrivilege(ProtectionDomain.java:74)
	at java.security.ProtectionDomain$JavaSecurityAccessImpl.doIntersectionPrivilege(ProtectionDomain.java:84)
	at java.awt.EventQueue$4.run(EventQueue.java:733)
	at java.awt.EventQueue$4.run(EventQueue.java:731)
	at java.security.AccessController.doPrivileged(Native Method)
	at java.security.ProtectionDomain$JavaSecurityAccessImpl.doIntersectionPrivilege(ProtectionDomain.java:74)
	at java.awt.EventQueue.dispatchEvent(EventQueue.java:730)
	at java.awt.EventDispatchThread.pumpOneEventForFilters(EventDispatchThread.java:205)
	at java.awt.EventDispatchThread.pumpEventsForFilter(EventDispatchThread.java:116)
	at java.awt.EventDispatchThread.pumpEventsForFilter(EventDispatchThread.java:109)
	at java.awt.WaitDispatchSupport$2.run(WaitDispatchSupport.java:190)
	at java.awt.WaitDispatchSupport$4.run(WaitDispatchSupport.java:235)
	at java.awt.WaitDispatchSupport$4.run(WaitDispatchSupport.java:233)
	at java.security.AccessController.doPrivileged(Native Method)
	at java.awt.WaitDispatchSupport.enter(WaitDispatchSupport.java:233)
	at java.awt.Dialog.show(Dialog.java:1084)
	at java.awt.Component.show(Component.java:1671)
	at java.awt.Component.setVisible(Component.java:1623)
	at java.awt.Window.setVisible(Window.java:1014)
	at java.awt.Dialog.setVisible(Dialog.java:1005)
	at net.sf.mzmine.parameters.impl.SimpleParameterSet.showSetupDialog(SimpleParameterSet.java:157)
	at net.sf.mzmine.modules.visualization.spectra.simplespectra.SpectraVisualizerWindow$9.run(SpectraVisualizerWindow.java:600)
	at java.awt.event.InvocationEvent.dispatch(InvocationEvent.java:311)
	at java.awt.EventQueue.dispatchEventImpl(EventQueue.java:758)
	at java.awt.EventQueue.access$500(EventQueue.java:97)
	at java.awt.EventQueue$3.run(EventQueue.java:709)
	at java.awt.EventQueue$3.run(EventQueue.java:703)
	at java.security.AccessController.doPrivileged(Native Method)
	at java.security.ProtectionDomain$JavaSecurityAccessImpl.doIntersectionPrivilege(ProtectionDomain.java:74)
	at java.awt.EventQueue.dispatchEvent(EventQueue.java:728)
	at java.awt.EventDispatchThread.pumpOneEventForFilters(EventDispatchThread.java:205)
	at java.awt.EventDispatchThread.pumpEventsForFilter(EventDispatchThread.java:116)
	at java.awt.EventDispatchThread.pumpEventsForHierarchy(EventDispatchThread.java:105)
	at java.awt.EventDispatchThread.pumpEvents(EventDispatchThread.java:101)
	at java.awt.EventDispatchThread.pumpEvents(EventDispatchThread.java:93)
	at java.awt.EventDispatchThread.run(EventDispatchThread.java:82)
```

**Comment by SteffenHeu** [comment]
Hi Tomas,
Unfortunately I couldn't reproduce the error with that parameters. I added an additional null check at that point which should fix the exception, although the object shouldn't be null in the first place since it's assigned in the constructor. 
Please let me know, if this fixed the error for you.

**Comment by tomas-pluskal** [comment]
I am still getting an exception with your latest code:
````
[15:47:10|FINEST|SpectraVisualizerWindow]: Loading scan #1 from BLANK sample for spectra visualizer
[15:47:32|FINEST|ProcessingComponent]: Added module 13C-Isotope grouper to processing list.
Exception in thread "AWT-EventQueue-0" java.lang.NullPointerException
	at net.sf.mzmine.modules.visualization.spectra.simplespectra.datapointprocessing.DataPointProcessingManager.clearProcessingSteps(DataPointProcessingManager.java:297)
	at net.sf.mzmine.parameters.parametertypes.ProcessingComponent.sendQueue(ProcessingComponent.java:299)
	at net.sf.mzmine.parameters.parametertypes.ProcessingComponent.actionPerformed(ProcessingComponent.java:91)
	at javax.swing.AbstractButton.fireActionPerformed(AbstractButton.java:2022)
	at javax.swing.AbstractButton$Handler.actionPerformed(AbstractButton.java:2348)
	at javax.swing.DefaultButtonModel.fireActionPerformed(DefaultButtonModel.java:402)
	at javax.swing.DefaultButtonModel.setPressed(DefaultButtonModel.java:259)
	at javax.swing.plaf.basic.BasicButtonListener.mouseReleased(BasicButtonListener.java:252)
	at java.awt.Component.processMouseEvent(Component.java:6539)
	at javax.swing.JComponent.processMouseEvent(JComponent.java:3324)
	at java.awt.Component.processEvent(Component.java:6304)
	at java.awt.Container.processEvent(Container.java:2239)
	at java.awt.Component.dispatchEventImpl(Component.java:4889)
	at java.awt.Container.dispatchEventImpl(Container.java:2297)
	at java.awt.Component.dispatchEvent(Component.java:4711)
	at java.awt.LightweightDispatcher.retargetMouseEvent(Container.java:4904)
	at java.awt.LightweightDispatcher.processMouseEvent(Container.java:4535)
	at java.awt.LightweightDispatcher.dispatchEvent(Container.java:4476)
	at java.awt.Container.dispatchEventImpl(Container.java:2283)
	at java.awt.Window.dispatchEventImpl(Window.java:2746)
	at java.awt.Component.dispatchEvent(Component.java:4711)
	at java.awt.EventQueue.dispatchEventImpl(EventQueue.java:760)
	at java.awt.EventQueue.access$500(EventQueue.java:97)
	at java.awt.EventQueue$3.run(EventQueue.java:709)
	at java.awt.EventQueue$3.run(EventQueue.java:703)
	at java.security.AccessController.doPrivileged(Native Method)
	at java.security.ProtectionDomain$JavaSecurityAccessImpl.doIntersectionPrivilege(ProtectionDomain.java:74)
	at java.security.ProtectionDomain$JavaSecurityAccessImpl.doIntersectionPrivilege(ProtectionDomain.java:84)
	at java.awt.EventQueue$4.run(EventQueue.java:733)
	at java.awt.EventQueue$4.run(EventQueue.java:731)
	at java.security.AccessController.doPrivileged(Native Method)
	at java.security.ProtectionDomain$JavaSecurityAccessImpl.doIntersectionPrivilege(ProtectionDomain.java:74)
	at java.awt.EventQueue.dispatchEvent(EventQueue.java:730)
	at java.awt.EventDispatchThread.pumpOneEventForFilters(EventDispatchThread.java:205)
	at java.awt.EventDispatchThread.pumpEventsForFilter(EventDispatchThread.java:116)
	at java.awt.EventDispatchThread.pumpEventsForFilter(EventDispatchThread.java:109)
	at java.awt.WaitDispatchSupport$2.run(WaitDispatchSupport.java:190)
	at java.awt.WaitDispatchSupport$4.run(WaitDispatchSupport.java:235)
	at java.awt.WaitDispatchSupport$4.run(WaitDispatchSupport.java:233)
	at java.security.AccessController.doPrivileged(Native Method)
	at java.awt.WaitDispatchSupport.enter(WaitDispatchSupport.java:233)
	at java.awt.Dialog.show(Dialog.java:1084)
	at java.awt.Component.show(Component.java:1671)
	at java.awt.Component.setVisible(Component.java:1623)
	at java.awt.Window.setVisible(Window.java:1014)
	at java.awt.Dialog.setVisible(Dialog.java:1005)
	at net.sf.mzmine.parameters.impl.SimpleParameterSet.showSetupDialog(SimpleParameterSet.java:157)
	at net.sf.mzmine.modules.visualization.spectra.simplespectra.SpectraVisualizerWindow$8.run(SpectraVisualizerWindow.java:590)
	at java.awt.event.InvocationEvent.dispatch(InvocationEvent.java:311)
	at java.awt.EventQueue.dispatchEventImpl(EventQueue.java:758)
	at java.awt.EventQueue.access$500(EventQueue.java:97)
	at java.awt.EventQueue$3.run(EventQueue.java:709)
	at java.awt.EventQueue$3.run(EventQueue.java:703)
	at java.security.AccessController.doPrivileged(Native Method)
	at java.security.ProtectionDomain$JavaSecurityAccessImpl.doIntersectionPrivilege(ProtectionDomain.java:74)
	at java.awt.EventQueue.dispatchEvent(EventQueue.java:728)
	at java.awt.EventDispatchThread.pumpOneEventForFilters(EventDispatchThread.java:205)
	at java.awt.EventDispatchThread.pumpEventsForFilter(EventDispatchThread.java:116)
	at java.awt.EventDispatchThread.pumpEventsForHierarchy(EventDispatchThread.java:105)
	at java.awt.EventDispatchThread.pumpEvents(EventDispatchThread.java:101)
	at java.awt.EventDispatchThread.pumpEvents(EventDispatchThread.java:93)
	at java.awt.EventDispatchThread.run(EventDispatchThread.java:82)
Exception in thread "AWT-EventQueue-0" java.lang.NullPointerException
	at net.sf.mzmine.modules.visualization.spectra.simplespectra.datapointprocessing.DataPointProcessingManager.clearProcessingSteps(DataPointProcessingManager.java:297)
	at net.sf.mzmine.parameters.parametertypes.ProcessingComponent.sendQueue(ProcessingComponent.java:299)
	at net.sf.mzmine.parameters.parametertypes.ProcessingComponent.setParameters(ProcessingComponent.java:196)
	at net.sf.mzmine.parameters.parametertypes.ProcessingComponent.actionPerformed(ProcessingComponent.java:96)
	at javax.swing.AbstractButton.fireActionPerformed(AbstractButton.java:2022)
	at javax.swing.AbstractButton$Handler.actionPerformed(AbstractButton.java:2348)
	at javax.swing.DefaultButtonModel.fireActionPerformed(DefaultButtonModel.java:402)
	at javax.swing.DefaultButtonModel.setPressed(DefaultButtonModel.java:259)
	at javax.swing.plaf.basic.BasicButtonListener.mouseReleased(BasicButtonListener.java:252)
	at java.awt.Component.processMouseEvent(Component.java:6539)
	at javax.swing.JComponent.processMouseEvent(JComponent.java:3324)
	at java.awt.Component.processEvent(Component.java:6304)
	at java.awt.Container.processEvent(Container.java:2239)
	at java.awt.Component.dispatchEventImpl(Component.java:4889)
	at java.awt.Container.dispatchEventImpl(Container.java:2297)
	at java.awt.Component.dispatchEvent(Component.java:4711)
	at java.awt.LightweightDispatcher.retargetMouseEvent(Container.java:4904)
	at java.awt.LightweightDispatcher.processMouseEvent(Container.java:4535)
	at java.awt.LightweightDispatcher.dispatchEvent(Container.java:4476)
	at java.awt.Container.dispatchEventImpl(Container.java:2283)
	at java.awt.Window.dispatchEventImpl(Window.java:2746)
	at java.awt.Component.dispatchEvent(Component.java:4711)
	at java.awt.EventQueue.dispatchEventImpl(EventQueue.java:760)
	at java.awt.EventQueue.access$500(EventQueue.java:97)
	at java.awt.EventQueue$3.run(EventQueue.java:709)
	at java.awt.EventQueue$3.run(EventQueue.java:703)
	at java.security.AccessController.doPrivileged(Native Method)
	at java.security.ProtectionDomain$JavaSecurityAccessImpl.doIntersectionPrivilege(ProtectionDomain.java:74)
	at java.security.ProtectionDomain$JavaSecurityAccessImpl.doIntersectionPrivilege(ProtectionDomain.java:84)
	at java.awt.EventQueue$4.run(EventQueue.java:733)
	at java.awt.EventQueue$4.run(EventQueue.java:731)
	at java.security.AccessController.doPrivileged(Native Method)
	at java.security.ProtectionDomain$JavaSecurityAccessImpl.doIntersectionPrivilege(ProtectionDomain.java:74)
	at java.awt.EventQueue.dispatchEvent(EventQueue.java:730)
	at java.awt.EventDispatchThread.pumpOneEventForFilters(EventDispatchThread.java:205)
	at java.awt.EventDispatchThread.pumpEventsForFilter(EventDispatchThread.java:116)
	at java.awt.EventDispatchThread.pumpEventsForFilter(EventDispatchThread.java:109)
	at java.awt.WaitDispatchSupport$2.run(WaitDispatchSupport.java:190)
	at java.awt.WaitDispatchSupport$4.run(WaitDispatchSupport.java:235)
	at java.awt.WaitDispatchSupport$4.run(WaitDispatchSupport.java:233)
	at java.security.AccessController.doPrivileged(Native Method)
	at java.awt.WaitDispatchSupport.enter(WaitDispatchSupport.java:233)
	at java.awt.Dialog.show(Dialog.java:1084)
	at java.awt.Component.show(Component.java:1671)
	at java.awt.Component.setVisible(Component.java:1623)
	at java.awt.Window.setVisible(Window.java:1014)
	at java.awt.Dialog.setVisible(Dialog.java:1005)
	at net.sf.mzmine.parameters.impl.SimpleParameterSet.showSetupDialog(SimpleParameterSet.java:157)
	at net.sf.mzmine.modules.visualization.spectra.simplespectra.SpectraVisualizerWindow$8.run(SpectraVisualizerWindow.java:590)
	at java.awt.event.InvocationEvent.dispatch(InvocationEvent.java:311)
	at java.awt.EventQueue.dispatchEventImpl(EventQueue.java:758)
	at java.awt.EventQueue.access$500(EventQueue.java:97)
	at java.awt.EventQueue$3.run(EventQueue.java:709)
	at java.awt.EventQueue$3.run(EventQueue.java:703)
	at java.security.AccessController.doPrivileged(Native Method)
	at java.security.ProtectionDomain$JavaSecurityAccessImpl.doIntersectionPrivilege(ProtectionDomain.java:74)
	at java.awt.EventQueue.dispatchEvent(EventQueue.java:728)
	at java.awt.EventDispatchThread.pumpOneEventForFilters(EventDispatchThread.java:205)
	at java.awt.EventDispatchThread.pumpEventsForFilter(EventDispatchThread.java:116)
	at java.awt.EventDispatchThread.pumpEventsForHierarchy(EventDispatchThread.java:105)
	at java.awt.EventDispatchThread.pumpEvents(EventDispatchThread.java:101)
	at java.awt.EventDispatchThread.pumpEvents(EventDispatchThread.java:93)
	at java.awt.EventDispatchThread.run(EventDispatchThread.java:82)


````

**Comment by tomas-pluskal** [comment]
You should ALWAYS use the Nullable and Nonnull annotations in all your methods. That will make it much easier to track such exceptions. For example, in this method:
```
public void setProcessingQueue(DataPointProcessingQueue list)
```
Can the list parameter be null or not? It's unclear.

**Comment by SteffenHeu** [comment]
I added nonnull and nullable annotations and an additional warning message. I could not reproduce the null pointer exception though. I also tried this on a Mac and did not get that error. If you still get that exception or one of this warnings:
`The processing queue is null and clearProcessingSteps was called.
The processing list was about to be set to null.`
please let me know.

**Comment by tomas-pluskal** [comment]
Ok, I don't get the exception anymore, instead I got this error:

Screen Shot 2019-04-01 at 12 59 30

In principle, I don't think it should be required to run the mass detection, because the idea is to process the "raw" spectra here, and if the spectra are already centroided, the mass detection does not add any useful information. But at least the module doesn't crash now :) After the error happened, there doesn't seem to be any way to restart the datapoint processing for this scan, though - even if I change the processing parameters and add the mass detection step, it does nothing. Only after loading a different scan it starts processing again.

**Comment by tomas-pluskal** [comment]
I think it's better to move the main checkbox to the bottom panel of the visualizer

Screen Shot 2019-04-01 at 13 05 16

Like this:
[ ] Spectrum processing {Configure}

Where the {Configure} button will be disabled unless the checkbox is checked.

**Comment by tomas-pluskal** [comment]
@SteffenHeu 
Do you have a time estimate for finishing this function? I would like to release a new MZmine version soon-ish.

**Comment by SteffenHeu** [comment]
Not really, sorry. I'm mid-semester right now so i don't have much time at the moment. I might be able to do some work this week, but you shouldn't wait for me to finish.

**Comment by tomas-pluskal** [comment]
Ok!

**Comment by SteffenHeu** [comment]
I implemented your latest suggestions, but it's not done yet. Still have to think about your suggestions regarding the use of existing parameter sets.

Also, I get this exception when running in normal/debug mode since I merged the latest version into this branch. I suppose this isn't normal?
```
Can't load log handler "java.util.logging.FileHandler"
java.nio.file.NoSuchFileException: log\mzmine.log.lck
java.nio.file.NoSuchFileException: log\mzmine.log.lck
	at sun.nio.fs.WindowsException.translateToIOException(WindowsException.java:79)
	at sun.nio.fs.WindowsException.rethrowAsIOException(WindowsException.java:97)
	at sun.nio.fs.WindowsException.rethrowAsIOException(WindowsException.java:102)
	at sun.nio.fs.WindowsFileSystemProvider.newFileChannel(WindowsFileSystemProvider.java:115)
	at java.nio.channels.FileChannel.open(FileChannel.java:287)
	at java.nio.channels.FileChannel.open(FileChannel.java:335)
	at java.util.logging.FileHandler.openFiles(FileHandler.java:459)
	at java.util.logging.FileHandler.<init>(FileHandler.java:263)
	at sun.reflect.NativeConstructorAccessorImpl.newInstance0(Native Method)
	at sun.reflect.NativeConstructorAccessorImpl.newInstance(NativeConstructorAccessorImpl.java:62)
	at sun.reflect.DelegatingConstructorAccessorImpl.newInstance(DelegatingConstructorAccessorImpl.java:45)
	at java.lang.reflect.Constructor.newInstance(Constructor.java:423)
	at java.lang.Class.newInstance(Class.java:442)
	at java.util.logging.LogManager$5.run(LogManager.java:966)
	at java.security.AccessController.doPrivileged(Native Method)
	at java.util.logging.LogManager.loadLoggerHandlers(LogManager.java:958)
	at java.util.logging.LogManager.initializeGlobalHandlers(LogManager.java:1578)
	at java.util.logging.LogManager.access$1500(LogManager.java:145)
	at java.util.logging.LogManager$RootLogger.accessCheckedHandlers(LogManager.java:1667)
	at java.util.logging.Logger.getHandlers(Logger.java:1777)
	at java.util.logging.Logger.log(Logger.java:735)
	at java.util.logging.Logger.doLog(Logger.java:765)
	at java.util.logging.Logger.log(Logger.java:788)
	at java.util.logging.Logger.info(Logger.java:1490)
	at net.sf.mzmine.main.MZmineCore.main(MZmineCore.java:91)
```

tomas-pluskal:
Hi @SteffenHeu 
The exception means that the log folder doesn't exist in your working directory. If you're using Eclipse, just set the working directory to src/main in your running configuration, as shown below:

Screen Shot 2019-05-26 at 13 48 09

tomas-pluskal:
Your code seems to have some compilation errors in ProcessingComponent.java (https://travis-ci.org/mzmine/mzmine2/jobs/537463510).

SteffenHeu:
Yeah i saw that, it's weird because it compiles fine on my PC and it did so before. That file was not changed during the last commits. I will have to check that.

tomas-pluskal:
Hi Steffen,
Do you have an ETA for this PR? I would like to release a new MZmine version in a few days. Or we can leave this PR for the version after.

SteffenHeu:
It depends. @robinschmid and @Annexhc have added a lot of new tools regarding the spectra identification. I would like to have a talk with them first, so this PR has all the functionality to be fit for their purposes.
Furthermore, you suggested changes regarding the task management, which i will have to think about. That would be a fundamental change to this PR and would take a lot of time, I think.

tomas-pluskal:
If the current code is fully functional and stable (e.g., doesn't generate exceptions or break the spectra visualizer in any way), we can merge it in current form as a prototype. Then the function can be improved later based on how we feel about using the prototype. It often takes a few iterations to arrive to the best design.

Annexhc:
@SteffenHeu I would love to talk to you about the new additions and how we can implement them into your processing module. However, I think it does not hurt anyone if we postpone the pull request one more time. There a quite a lot of new additions for the next release anyway. Also, I think it would be great to release 2.40 prior to ASMS. Nevertheless, I totally agree with Tomas, that if your branch is stable we should merge it.

robinschmid:
Same here. The last time I tried it, the whole thing was very user-friendly and round. There are lots of methods which need to be added in the future - and Ansgar and I will have to change some of our methods to add them to this processing scheme - but I would suggest that you say if the prototype is ready to be merged into the next version.

If you feel rushed, we can merge it after releasing the new version. Otherwise, lets merge and create a new version. 

Cheers,
Robin

SteffenHeu:
I will do some minor additions/changes today and during the weekend. If no big issues come up, I should be able to finish this in time. I will try to do some more intense testing though.

tomas-pluskal:
Sounds good. I would like to release the next version by Sunday, before the ASMS starts on Monday.

tomas-pluskal:
@SteffenHeu Please remove the ```conf/config.xml``` file from this PR, thanks.

SteffenHeu:
Currently the processing is disbabled by default for every instance of SpectraPlot (can be enabled in constructor) but enabled for every instance of SpectraVisualizerWindow. That way it is not executed in preview windows. 
Is there some module I'm missing, that also uses SpectraVisualizerWindow but should not execute the processing?

tomas-pluskal:
I am not aware of any such module, but you could try searching the whole repository for “SpectraVisualizerWindow”.

Would you say it’s ready for merging now?

SteffenHeu:
I found some other references in:
net.sf.mzmine.modules.peaklistmethods.io.spectraldbsubmit.view.ScanSelectPanel.createChart()
net.sf.mzmine.modules.visualization.spectra.simplespectra.MultiSpectraVisualizerWindow.addSpectra(int)
net.sf.mzmine.util.dialogs.FeatureOverviewWindow.addSpectraMS1()
net.sf.mzmine.util.dialogs.FeatureOverviewWindow.addSpectraMS2()
so i added another constructor to the SpectraVisualizerWindow, so it is disabled by default.
Also, unticking the checkbox for the processing will reload the current scan and remove the processing results, so you don't have to switch scans anymore.

I couldn't find any more bugs, but of course I know how to use this module and might not run into more common mistakes, but i tried to test everything. So if you are happy with how it is implemented now, I am fine with merging it.
If you find any mistakes just let me know!

robinschmid:
The second consturctor which defaults to deactivated processing is good

tomas-pluskal:
Looks good to me. I will merge it and release new version. If we find additional bugs or problems (and we will), we can release a follow-up update relatively soon.

PR #547, opened 2019-03-11
Add code for calculating Student's t-test and logarithm of the fold change

asmirn1:
Hi Tomas,

This module calculates the following statistics for each aligned peak/feature:
- Student's t-test for two groups and its p-value,
- Logarithm of the fold-change for two groups.

Please, see if you would be interested in merging these changes.


tomas-pluskal:
Looks good, and doesn't break any existing functions, so I don't see any problem.

PR #549, opened 2019-03-15
Spectral MS/MS library submit and search

robinschmid:
Hey Tomas,

This PR contains two new modules.
The first adds a library submission window to the export menu of the peaklist. The user can select multiple ion/adducts of the same compound, select the best filtered MS/MS scan, provide meta data and create spectral library entries which can be saved to a local json file or directly submitted to the GNPS library. 

The second module allows to import a local spectral library (json) and to search for matches in a peaklist. In the future we should also develop an import/export for other formats (e.g., the massbank or NIST format). Users can visualise the database matches in a mirror plot.

I still need to create a video tutorial and some help files for this one.

Cheers
Robin

image
image


tomas-pluskal:
Hi Robin, this is a fantastic addition, thanks!
I will play with the module a bit and give you some feedback.

tomas-pluskal:
A few comments for spectral search:
- the field "Database file" doesn't give any information about what format the file is. Better name it "Database file (JSON)" and add a detailed description of the file format to the Help.
- "Noise level" seems redundant, because noise is already filtered during mass detection. What is the reasoning for this parameter?
- If I try to load a database with a different extension than .json (e.g., .txt), it automatically adds the .json extension and then it crashes with unhandled exception. This is wrong, just use the filename as provided.
- I got this exception when I used a .json file downloaded from MoNA:
Screen Shot 2019-03-16 at 12 21 26

tomas-pluskal:
A few comments for spectral export:
- again, the "Export MS/MS library" button should perhaps be renamed to "Export MS/MS library (JSON)"
- the mass list name should be specified using MassListParameter
- again, I think noise level is redundant
- the Submit to GNPS should be an OptionalModuleParameter with all the other settings specified in its own ParameterSet 
- the "From scan" button seems to do nothing for me
- I haven't been able to export anything, actually - it keeps saying "NO MS2 SPECTRA", even I selected a peak list row with nice MS2 data

robinschmid:
Hey Tomas

thanks for your comments. 
Did you try an old project? The export window focusses on the lately introduced list of all MS/MS scans (by @Annexhc ). Old projects still only have the "best" MS/MS scan.
Maybe we need another method (apart from the deconvolution) to assing all MS/MS scans to features in an aligned peaklist. 

This PR was a bit quick to show some preliminary developments and get some ideas from your side.
So far, the library matching is only possible against the json library that is generated by the MS/MS library export. This format was chosen based on the GNPS library submission format.

The next step inlcude the massbank json format and the msp format.

I will take care of the other comments in the next commits.

Cheers,
Robin

tomas-pluskal:
Hi Robin, @Annexhc 
I thought that when an old project is imported, MZmine would automatically rebuild the list of all MS/MS scans... it doesn't? All the information to build the list is included in the old project as well, isn't it?

Annexhc:
Hi Tomas,
no it does not. An old project needs to be reprocessed with MZmine 2.38 or newer. The import function only reads out the xml file and does not call the methods to rebuild the feature lists.

tomas-pluskal:
Hmm, but when the features are imported (during project import), why not update each feature with the correct MS/MS lists? Then storing the MS/MS lists won't even be necessary. Maybe I'm missing something here, though...

Annexhc:
Well, I think we could change that, but I need some time to think about this. If features lists are updated while reloading a project we don't need to store things like the isotope pattern as well. On the other side there might be some performance issues. Reloading or reprocessing is slower compared to just reading out the xml file.

tomas-pluskal:
I think is definitely better to rebuild this information when the features are loaded. It will preserve backwards compatibility of the project files. The performance issues can be mitigated by preloading some data into memory (e.g., a list of all MS/MS scans, their retention times, and precursor m/z values). Then the data can be added from this list stored in memory.

robinschmid:
We would then only need to store the pairing parameters. Retention time window and mz range.

tomas-pluskal:
I believe the RT and m/z range were already included in the feature XML description in the project in older MZmine versions.

tomas-pluskal:
I just realized you guys added these options to the Chromatogram deconvolution module:

Screen Shot 2019-04-02 at 14 48 19

It makes sense that the MS2 list needs to be saved as part of the project file. So let's keep things as they are, I think we should just reprocess and update the sample project file.

It would be great to finish this PR so we can release the next MZmine version soon!

robinschmid:
Yeah. I will try to work on this one to finish it. 
These options were added a while ago by some students of the dorrestein lab. I thought maybe it is better to change the mz pairing range to an mzrange parameter....
What do you think?

tomas-pluskal:
I agree.

Tomas

tomas-pluskal:
@robinschmid Shall we release v. 2.39 after this PR is finished? I think the other pending PR by @SteffenHeu will take a bit more time.

robinschmid:
Yeah this would be great. 
I will work on this next wednesday.

Lets create a todo list
 - mz and rt pairing parameters to tolerances
 - massbank (mona) format import and export
 - compatibility for old projects in terms of MS 2 list (what is the best idea?)


Compatibility:
We could just offer a new MS2 scan pairing module. People could run this module on chromatograms, peaklists and aligned lists...
I think this should be easier than rerunning the MS2 pairing.

tomas-pluskal:
I think it is fine to say that the data needs to be reprocessed to get the MS2 pairing right. I will make a note in the Release Notes for 2.39 about that. Let's invest time into more useful things than maintaining backwards compatibility :)

robinschmid:
Here is a first update.
You can import MONA json or GNPS' internal json format. 

You can give it a try: 
http://mona.fiehnlab.ucdavis.edu/downloads

At this stage, the library matching is very basic and we should implement some different matching algorithms.

tomas-pluskal:
Thanks, Robin!
I think it would be good to merge this PR at this point, and leave further updates for next MZmine versions. It would be great if you could add some help files to these new modules, at least describing the basic functions (I understand that the modules might still change later).

robinschmid:
Let me check the modules again tomorrow. I am happy to merge them at this point.

robinschmid:
All done and checked. I have also added some help files.

Additions are:
- spectral matching against local database
  - msp NIST format
  - json/msp MoNA format (tested the whole MoNA LC-positive on it)
  - json GNPS internal library submission format (also generated from library submission module)
- MS2 library submission module
  - submission of multiple adducts, in-source fragments etc at once to...
  - ...GNPS online
  - ... json and msp local library
- MS2 scan to feature grouping
  - a small module that groups all MS2 scans to their feature
  - most people should run this module on their older projects (<2.38)
  - the module is located in peaklist/filters (I had no clue where to put it)

From my side this PR is ready for merging.

Cheers
Robin

robinschmid:
Tested the spectral matching with 65000 entries in a MoNA json and it took 7 minutes on my laptop.

tomas-pluskal:
Hey Robin, this looks like a fantastic addition to MZmine! 
I'll take a closer look at the code tomorrow and if everything looks fine, I'll merge it.

src/main/java/net/sf/mzmine/modules/peaklistmethods/filtering/groupms2/GroupMS2Task.java
```diff
@@ -0,0 +1,164 @@
...
+              if ((!limitRTByFeature || rtRange.contains(scan.getRetentionTime()))
+                  && rtTol.checkWithinTolerance(frt, scan.getRetentionTime())) {
+                if (scan.getPrecursorMZ() != 0
+                    && mzTol.checkWithinTolerance(fmz, scan.getPrecursorMZ()))
+                  scans.add(i);
+              } else {
+                // end of loop - out of tolerance
+                break;
+              }
+            } catch (Exception e) {
```
tomas-pluskal: At least add e.printStackTrace(), you should never swallow exceptions without warning.

src/main/java/net/sf/mzmine/modules/peaklistmethods/filtering/groupms2/GroupMS2Task.java
```diff
@@ -0,0 +1,164 @@
...
+          RawDataFile raw = f.getDataFile();
+          IntArrayList scans = new IntArrayList();
+          int best = f.getRepresentativeScanNumber();
+          double frt = f.getRT();
+          double fmz = f.getMZ();
+          Range<Double> rtRange = f.getRawDataPointsRTRange();
+          int i = best;
+          // left
+          while (i > 1) {
+            try {
```
tomas-pluskal: This kind of ````try {} catch (Exception e) {}```` block looks fishy. What kind of exception are you expecting? Maybe clarify in a comment.

src/main/java/net/sf/mzmine/modules/peaklistmethods/filtering/groupms2/help/help.html
```diff
@@ -0,0 +1,28 @@
...
+		<link rel="stylesheet" type="text/css" href="/net/sf/mzmine/desktop/impl/helpsystem/HelpStyles.css">
+    </head>
+
+<body>
+
+<h1>MS2 to feature grouping</h1>
+
+<h2>Description</h2>
+<p>
+Call this method on any (aligned) feature list to sort all fragmention scans to their corresponding features.
```
tomas-pluskal: Change "sort" to "assign"?

src/main/java/net/sf/mzmine/modules/peaklistmethods/identification/spectraldbsearch/LocalSpectralDBSearchParameters.java
```diff
@@ -0,0 +1,64 @@
...
+      "(GNPS json, MONA json, NIST msp) Name of file that contains information for peak identification");
+
+  public static final MassListParameter massList =
+      new MassListParameter("MassList (MS2)", "MassList for MS/MS scans to match");
+
+  public static final OptionalParameter<RTToleranceParameter> rtTolerance =
+      new OptionalParameter<>(new RTToleranceParameter());
+
+  public static final MZToleranceParameter mzTolerance = new MZToleranceParameter();
+  public static final DoubleParameter noiseLevelMS2 =
```
tomas-pluskal: Why do you need this noise level? There is already a noise level parameter in the mass detection module.

src/main/java/net/sf/mzmine/modules/peaklistmethods/identification/spectraldbsearch/LocalSpectralDBSearchParameters.java
```diff
@@ -0,0 +1,64 @@
...
+      new DoubleParameter("Noise level", "Noise level to filter MS2 mass lists",
+          MZmineCore.getConfiguration().getIntensityFormat(), 0d);
+  public static final DoubleParameter minCosine = new DoubleParameter("Min cos similarity",
+      "Cosine similarity", new DecimalFormat("0.000"), 0.7);
+  public static final IntegerParameter minMatch = new IntegerParameter("Min matched signals",
```
tomas-pluskal: What are "signals"?

src/main/java/net/sf/mzmine/modules/peaklistmethods/identification/spectraldbsearch/LocalSpectralDBSearchParameters.java
```diff
@@ -0,0 +1,64 @@
...
+
+  public static final OptionalParameter<RTToleranceParameter> rtTolerance =
+      new OptionalParameter<>(new RTToleranceParameter());
+
+  public static final MZToleranceParameter mzTolerance = new MZToleranceParameter();
+  public static final DoubleParameter noiseLevelMS2 =
+      new DoubleParameter("Noise level", "Noise level to filter MS2 mass lists",
+          MZmineCore.getConfiguration().getIntensityFormat(), 0d);
+  public static final DoubleParameter minCosine = new DoubleParameter("Min cos similarity",
+      "Cosine similarity", new DecimalFormat("0.000"), 0.7);
```
tomas-pluskal: The "Cosine similarity" description could be a bit improved.

src/main/java/net/sf/mzmine/modules/peaklistmethods/io/spectraldbsubmit/LibrarySubmitParameters.java
```diff
@@ -0,0 +1,177 @@
...
+    private IonSource(String value) {
+      this.value = value;
+    }
+
+    @Override
+    public String toString() {
+      return value != null ? value : super.toString();
+    }
+  }
+  public enum Instrument {
```
tomas-pluskal: I would suggest putting this enum into a separate file, for clarity

src/main/java/net/sf/mzmine/modules/peaklistmethods/io/spectraldbsubmit/LibrarySubmitParameters.java
```diff
@@ -0,0 +1,177 @@
...
+
+  public enum CompoundSource {
+  Lysate, Isolated, Commercial, Crude, Other;
+  }
+
+  public enum Polarity {
+    Positive, Negative;
+  }
+
+  public enum IonSource {
```
tomas-pluskal: I would suggest putting this enum into a separate file, for clarity

src/main/java/net/sf/mzmine/modules/peaklistmethods/io/spectraldbsubmit/LibrarySubmitTask.java
```diff
@@ -0,0 +1,241 @@
...
+      fileMSP = exportMSPFile ? FileAndPathUtil.getRealFilePath(tmpfile, "msp") : null;
+    } else {
+      fileJson = null;
+      fileMSP = null;
+    }
+  }
+
+  @Override
+  public double getFinishedPercentage() {
+    return (done / map.size());
```
tomas-pluskal: What if map.size() == 0?

src/main/java/net/sf/mzmine/modules/peaklistmethods/io/spectraldbsubmit/LibrarySubmitTask.java
```diff
@@ -0,0 +1,241 @@
...
+
+        // ######################################################
+        // NEEDED
+        // user pass and json entry
+        entity.addPart("username", new StringBody(USER));
+        entity.addPart("password", new StringBody(PASS));
+        entity.addPart("spectrum", new StringBody(json));
+
+        HttpPost httppost =
+            new HttpPost("http://dorresteinappshub.ucsd.edu:5050/depostsinglespectrum");
```
tomas-pluskal: Move the URL to a private static final String field

src/main/java/net/sf/mzmine/modules/peaklistmethods/io/spectraldbsubmit/sorting/ScanSorter.java
```diff
@@ -0,0 +1,45 @@
...
+
+package net.sf.mzmine.modules.peaklistmethods.io.spectraldbsubmit.sorting;
+
+import java.util.Comparator;
+import net.sf.mzmine.datamodel.MassList;
+import net.sf.mzmine.datamodel.Scan;
+import net.sf.mzmine.util.ScanUtils;
+import net.sf.mzmine.util.exceptions.MissingMassListException;
+
+public class ScanSorter implements Comparator<Scan> {
```
tomas-pluskal: This class can be of general use, so you should move it to the net.sf.mzmine.util package

src/main/java/net/sf/mzmine/util/ScanUtils.java
```diff
@@ -603,4 +612,118 @@ public static int findBestFragmentScan(RawDataFile dataFile, Range<Double> rtRan
...
+   * 
+   * @param scan
+   * @param massListName
+   * @return null if no masslist with this name or if name was not specified and this scan has zero
+   *         masslists
+   * @throws MissingMassListException
+   */
+  public static MassList getMassListOrFirst(Scan scan, String massListName) {
+    final MassList massList;
+    if (massListName == null || massListName.length() == 0) {
```
tomas-pluskal: Just FYI, there is a convenient ````Strings.isNullOrEmpty()```` method in Google Guava

src/main/java/net/sf/mzmine/modules/peaklistmethods/io/spectraldbsubmit/sorting/MassListSorter.java
```diff
@@ -0,0 +1,68 @@
...
+ * USA
+ */
+
+package net.sf.mzmine.modules.peaklistmethods.io.spectraldbsubmit.sorting;
+
+import java.util.Comparator;
+import net.sf.mzmine.datamodel.DataPoint;
+import net.sf.mzmine.util.ScanUtils;
+
+public class MassListSorter implements Comparator<DataPoint[]> {
```
tomas-pluskal: This class can also be moved to net.sf.mzmine.util package

src/main/java/net/sf/mzmine/modules/peaklistmethods/io/spectraldbsubmit/sorting/MassListSorter.java
```diff
@@ -0,0 +1,68 @@
...
+  private double noiseLevel;
+  private ScanSortMode sort;
+
+  public MassListSorter(ScanSortMode sort, double noiseLevel) {
```
tomas-pluskal: Why is the noise level necessary here? Mass lists are already filtered. Adding another noise level parameter just creates confusion for users.

tomas-pluskal:
Hey Robin,
I added a few minor comments to your code, please take a look.

src/main/java/net/sf/mzmine/modules/peaklistmethods/identification/spectraldbsearch/LocalSpectralDBSearchParameters.java
robinschmid: I just like to have an additional noise level here. The user does not have to run MassDetection again to use a higher noise level. (Convenience)

src/main/java/net/sf/mzmine/modules/peaklistmethods/filtering/groupms2/GroupMS2Task.java
robinschmid: Yeah changed it to a null checking. 
The RawDataFile.getScan(i) method should be @Nullable 
The user can enter any index which is out of range or just not in the HashTable (the scan numbers do not have to be contiguous).

src/main/java/net/sf/mzmine/modules/peaklistmethods/identification/spectraldbsearch/LocalSpectralDBSearchParameters.java
robinschmid: Feel free to change it in any way. 
The description is changed to:
Minimum number of matched signals in MS2 masslist and spectral library entry (within mz tolerance)

src/main/java/net/sf/mzmine/modules/peaklistmethods/identification/spectraldbsearch/LocalSpectralDBSearchParameters.java
robinschmid: Added a few more details. 
We still need a better checking method. Maybe the one from MassBank?

src/main/java/net/sf/mzmine/modules/peaklistmethods/io/spectraldbsubmit/LibrarySubmitParameters.java
robinschmid: All enums are now located in a separate class (GnpsValues) to link them to the specific subject
this allows GnpsValues.Instrument. ...

Is that ok?

src/main/java/net/sf/mzmine/modules/peaklistmethods/io/spectraldbsubmit/sorting/MassListSorter.java
robinschmid: The noise level is a convenience method. The user does not have to rerun massdetection. 
Especially helpful in the library submit module. The user should take a look at the spectrum and define a resonable noise level to assure the quality of the spectral library entries.

src/main/java/net/sf/mzmine/modules/peaklistmethods/io/spectraldbsubmit/sorting/ScanSorter.java
robinschmid: done

src/main/java/net/sf/mzmine/util/ScanUtils.java
robinschmid: I have moved this class to util.scans

robinschmid:
I have tried to answer and work on all your comments. Thanks for the review.

tomas-pluskal:
Thanks, Robin!
About the noise level: okay, but in this case I would prefer to call it something different, like "Min ion intensity" and in the Help we should explain why this parameter exists in addition to the "Noise level" in Mass detection. What do you think about that?

Issue #552, opened 2019-03-28
Isotope peak scanner issues


tomas-pluskal:
The isotope peak scanner module executed with these parameters:

<img width="739" alt="Screen Shot 2019-03-28 at 11 38 24" src="https://user-images.githubusercontent.com/1796692/55171342-87411880-514e-11e9-9a24-edeb6af7cd2b.png">

produces this error:

<img width="689" alt="Screen Shot 2019-03-28 at 11 38 36" src="https://user-images.githubusercontent.com/1796692/55171353-8c9e6300-514e-11e9-9f55-587fa4936f07.png">

Also, in the task progress window, it shows as "Got parameters" with status CANCELED (should be ERROR).

<img width="1165" alt="Screen Shot 2019-03-28 at 11 42 53" src="https://user-images.githubusercontent.com/1796692/55171425-b6f02080-514e-11e9-8171-73e0cd3841e6.png">

tomas-pluskal:
@robinschmid Could you please take a look at this?

robinschmid:
Yeah will do.

PR #554, opened 2019-03-30
Isotope peak scanner - fix for issue #552

SteffenHeu:
Hi Tomas,
this PR adresses the issue #552, thanks for reporting it. If you notice anything else just let me know.

Best regards
Steffen

### DIFF
--- a/src/main/java/net/sf/mzmine/modules/peaklistmethods/isotopes/isotopepeakscanner/IsotopePeakScannerTask.java
+++ b/src/main/java/net/sf/mzmine/modules/peaklistmethods/isotopes/isotopepeakscanner/IsotopePeakScannerTask.java
@@ -748,31 +748,27 @@ public class IsotopePeakScannerTask extends AbstractTask {
 
   private boolean checkParameters() {
     if (charge == 0) {
+      setErrorMessage("Error: charge may not be 0!");
       setStatus(TaskStatus.ERROR);
-      MZmineCore.getDesktop().displayMessage(MZmineCore.getDesktop().getMainWindow(),
-          "Error: charge may not be 0!");
       return false;
     }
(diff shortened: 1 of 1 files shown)


src/main/java/net/sf/mzmine/modules/peaklistmethods/isotopes/isotopepeakscanner/IsotopePeakScannerTask.java
```diff
@@ -404,7 +404,8 @@ public void run() {
         // The results you miss by skipping here would have not been valid results anyway, so this
         // is not urgent. Will be nicer though, because of cleaner code.
 
-      PeakListRow parent = copyPeakRow(peakList.getRow(i));
+//      PeakListRow parent = copyPeakRow(peakList.getRow(i));
+      PeakListRow parent = copyPeakRow(plh.getRowByID(candidates[bestPatternIndex].get(0).getCandID()));
```
tomas-pluskal: This kind of statement is bad practice, because it contains multiple points where it may throw a NullPointerException or ArrayIndexOutOfBoundsException. Please split this into multiple lines, with each line performing a single operation. Also, add proper checks that the variables don't have any unexpected values (like null).

SteffenHeu: Done. Generally, the results are checked before, if they are ok, else it doesn't get to this point.
But I also think, that the code of module is a bit messy. After i resolved the DataPointProcessing stuff, my goal is to rewrite this and make it cleaner and easier understandable and more functional.

tomas-pluskal:
Thanks!

Issue #553, opened 2019-03-29
Setting the m/z value of a detected peak in MZmine v2.36 and MZmine v2.38


asmirn1:
Hi Tomas,

I'm not reporting a bug, but I have one suggestion.

In MZmine v2.36, the m/z value of a detected peak was the same as that of the chromatogram where that peak was detected. So, all peaks detected in one chromatogram would have the same m/z value. In MZmine v2.38, the m/z value of a detected peak is calculated individually for each peak. So, it's hard to tell which chromatogram a peak was detected in (See the attached figure).

In our work, the behavior of MZmine v2.36 is preferable, but it cannot be replicated in MZmine v2.38. I wonder, if we could add an option for parameter *m/z center calibration* in the Chromatogram Deconvolution, that would allow setting the m/z value of a detected peak to the m/z value of the corresponding chromatogram, as it was in MZmine v2.36.

I could work on adding that option.

![version_comparison](https://user-images.githubusercontent.com/13631120/55243217-216b9400-5215-11e9-9fff-132bd26ed098.png)

### Setup

MZmine v2.36 and MZmine v2.38 on MacOSX High Sierra.

tomas-pluskal:
Hi Alex, 

I cannot support such change, because it is conceptually wrong. The m/z value of each peak should be defined by the data points of that peak, not by the process by which a chromatogram was constructed.

Perhaps if you explain what is your goal, we can find a more systematic solution.

Tomas

asmirn1:
My goal is to check if two peaks are produced from the same chromatogram. In the attached figure, I have two peaks detected in the same chromatogram. You can see that those detected peaks are two parts of a single composite peak. When I'm doing spectral deconvolution, I get better results if I merge those two parts together. To handle such cases, I need to know when two peaks are detected in the same chromatogram. So I was using m/z values of chromatograms and detected peaks for that purpose. 

I understand that using m/z values is not the best solution. If you can think about a better one, please let me know.

![splitting](https://user-images.githubusercontent.com/13631120/55246448-9a6dea00-521b-11e9-937e-2a0f542f8066.png)

I agree that using the m/z value of a chromatogram for detected peaks may be conceptually wrong, but it was the default behavior of MZmine up until version 2.36, so maybe we could add that option for *m/z center calculation* for compatibility.

asmirn1:
As an alternative solution, we could add a variable to class `Feature`, so that a feature would contain some sort of a reference to the feature it was made from.

[...]

PR #555, opened 2019-04-01
Parent row ID that we discussed in #553

asmirn1:
Hi Tomas,

These are changes to store ID of a parent row. A few comments about my changes:

1. As I understood our discussion in #553, we have decided to use the field ID of PeakListRow as an identifier of a parent feature (chromatogram). In this case, I believe that it makes sense to store the parent ID as a property of PeakListRow instead of Feature. Let me know if you want to change it.

2. I've noticed that there are several versions of ProjectLoad... methods. I've changed the latest version (2_5). I'm not sure if I should change the older versions or make a new version for my changes?

3. I'll push changes to spectral deconvolution that are using this new feature in a separate PL.

Alex.

### DIFF
--- a/src/main/java/net/sf/mzmine/datamodel/impl/SimplePeakListRow.java
+++ b/src/main/java/net/sf/mzmine/datamodel/impl/SimplePeakListRow.java
@@ -38,8 +38,6 @@ import net.sf.mzmine.util.PeakSorter;
 import net.sf.mzmine.util.SortingDirection;
 import net.sf.mzmine.util.SortingProperty;
 
-import javax.annotation.Nullable;
-
 /**
  * Implementation of PeakListRow
  */
@@ -54,9 +52,6 @@ public class SimplePeakListRow implements PeakListRow {
   private PeakInformation information;
   private int myID;
   private double maxDataPointIntensity = 0;
-
-  @Nullable
-  private Integer parentRowID;
 
   /**
    * These variables are used for caching the average values, so we don't need to calculate them
@@ -416,17 +411,5 @@ public class SimplePeakListRow implements PeakListRow {
     this.calculateAverageValues();
   }
   // End Gauthier edit
-
-
-  @Override
-  public void setParentRowID(@Nullable Integer id) {
-    parentRowID = id;
-  }
-
-  @Override
-  @Nullable
-  public Integer getParentRowID() {
-    return parentRowID;
-  }
 }
 // End DorresteinLab edit
--- a/src/main/java/net/sf/mzmine/datamodel/impl/SimplePeakListRow.java
+++ b/src/main/java/net/sf/mzmine/datamodel/impl/SimplePeakListRow.java
@@ -38,6 +38,8 @@ import net.sf.mzmine.util.PeakSorter;
 import net.sf.mzmine.util.SortingDirection;
 import net.sf.mzmine.util.SortingProperty;
 
+import javax.annotation.Nullable;
+
 /**
  * Implementation of PeakListRow
  */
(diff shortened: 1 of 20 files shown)


tomas-pluskal:
Hi Alex,

The parent chromatogram ID should be a property of Feature, otherwise you will create a mess after alignment.

Updating the latest ProjectLoad.. module is fine.

Since you added this parent ID, it would be nice if the user has a way to see it somehow. E.g., you can add it to the peak list table visualizer as a new DataFileColumnType.

Please also make sure the parent ID is correctly retained after deisotoping, alignment, gap filling, etc.

src/main/java/net/sf/mzmine/datamodel/impl/SimplePeakListRow.java
```diff
@@ -52,6 +54,7 @@
   private PeakInformation information;
   private int myID;
   private double maxDataPointIntensity = 0;
+  private Integer parentRowID;
```
tomas-pluskal: Please use the ```@Nullable``` annotation

src/main/java/net/sf/mzmine/datamodel/impl/SimplePeakListRow.java
```diff
@@ -413,5 +416,15 @@ public void update() {
   // End Gauthier edit
 
 
+  @Override
+  public void setParentRowID(Integer id) {
```
tomas-pluskal: Please use the ```@Nullable``` annotation

asmirn1:
Hi Tomas,

Regarding DataFileColumnType. I see how you create columns in class `PeakListTable`, but I couldn't find the place where table cells are filled out with actual values from a peaklist. Could you tell me what class I need to change to fill my new column with values?

tomas-pluskal:
Hi Alex,
check the ```PeakListTableModel.getValueAt()``` method.

asmirn1:
Hi Tomas,

> Please also make sure the parent ID is correctly retained after deisotoping, alignment, gap filling, etc

If we continue this list of modules, it may include almost all peaklist methods in MZmine. I don't think it's a good idea to change all of them. So before I'll start making changes, could we come up with a short list of modules that need to be changed?

My original idea was to make the minimum number of changes so that:
1. The ADAP workflow (i.e. **Chromatogram builder** -> **Chromatogram deconvolution** -> **Spectral deconvolution** -> **Alignment**) would work as it is supposed to work, and
2. My changes won't break any other modules, algorithms, and workflows.

Therefore, I was planning to change only **Chromatogram deconvolution** module to save the chromatogram ID and **Spectral deconvolution** to read that ID. 

I will also make sure that **Spectral deconvolution** works even when the chromatogram ID is missing (i.e. `null`). So if users don't follow the ADAP workflow, they'll still be able to do spectral deconvolution, but they will not have that feature of merging adjacent peaks from the same chromatogram.

This way I will avoid changing everybody's code and potentially breaking something.

If this doesn't work for you and you would like to have changes in other modules, let's make a list of such modules.

tomas-pluskal:
Hi Alex,

I think most of the downstream modules (like deisotoping) create instances of SimpleFeature with a copy constructor:

https://github.com/mzmine/mzmine2/blob/master/src/main/java/net/sf/mzmine/modules/peaklistmethods/isotopes/deisotoper/IsotopeGrouperTask.java#L209

So all you need to do is to modify the copy constructor in SimpleFeature and make sure the parent row ID gets propagated to the new instance.

asmirn1:
Okay. I'll try that.

asmirn1:
Hi Tomas,

I've made the following changes:

- Remove ParentRowID from PeakListRow.
- Add ParentChromatographyRowID to SimpleFeature and ResolvedPeak
- Replace Feature[] with ResolvedPeak[] in the deconvolution methods to be able to use method setParentChromatographyRowID() of ResolvedPeak

Could you take a look at my changes.

Alex.

tomas-pluskal:
Hi Alex, 
Looks good, thanks!

asmirn1:
Thank you Tomas! 

I'll make PR with corresponding changes to the spectral deconvolution soon.

Issue #556, opened 2019-04-01
Neutral loss filtering error


bjonnh:
### Setup

I am reporting a problem with MZmine 2 version 2.38.
My operating system is a Linux 64.

Below I pasted the exact output of the following commands:
```
java -version
openjdk version "11.0.3" 2019-04-16
OpenJDK Runtime Environment (build 11.0.3+4)
OpenJDK 64-Bit Server VM (build 11.0.3+4, mixed mode)
```

### Expected behavior

When doing a filtering of a peak-list, I would like to be able to search for a specific neutral loss by value OR  molecular formula. 

### Actual behavior

First, it is not possible to put either value or the molecule, the window asks for both (but the help says that if a molecule is set, then value is not taken into account).

Adding a formula and a mass, it then crash:

16:34:32|FINEST|MainMenu]: Starting module Neutral loss filter with parameters Peak lists: 215_MSMS_1-35_01_3004.mzXML filtered chromatogramsADAP deconvoluted deisotoped, m/z tolerance: 0.01 m/z or 5.0 ppm, Check RT: true, Retention time tolerance: 0.1 min, Minimum height: 100.0, Neutral loss (m/z): 16.02, Molecule: NH2, Name suffix: auto
[16:34:32|FINEST|TaskQueue]: Adding task "Got paramenters..." to the task controller queue
[16:34:32|INFO|WorkerThread]: Starting processing of task Got paramenters...
[16:34:32|SEVERE|WorkerThread]: Unhandled exception java.lang.NullPointerException while processing task Got paramenters...(java.lang.NullPointerException@net.sf.mzmine.modules.peaklistmethods.filtering.neutralloss.NeutralLossFilterTask.setUpDiff(NeutralLossFilterTask.java:292))
java.lang.NullPointerException
	at net.sf.mzmine.modules.peaklistmethods.filtering.neutralloss.NeutralLossFilterTask.setUpDiff(NeutralLossFilterTask.java:292)
	at net.sf.mzmine.modules.peaklistmethods.filtering.neutralloss.NeutralLossFilterTask.run(NeutralLossFilterTask.java:116)
	at net.sf.mzmine.taskcontrol.impl.WorkerThread.run(WorkerThread.java:58)

### Steps to reproduce the problem

Create a peaklist.
Go to filtering, Neutral loss filter. enter parameters :
[...]

PR #557, opened 2019-04-02
Fix: Formula in neutral loss filter

robinschmid:
This is a fix for the issue #556 

It only includes the correct calculation of exact masses from string formulae. It also provides checking for correct formulae.


tomas-pluskal:
Thanks, Robin!

PR #558, opened 2019-04-02
fixed Neutral loss filter assignments

SteffenHeu:
Hi Tomas,
i fixed the wrong assignments robin mentioned in #556. It happened when the peak list was not sorted by ascending m/z. This issue is also occured in the isotope peak scanner, but is fixed already. It's using the ids for everything now.


tomas-pluskal:
Awesome, thanks!

PR #560, opened 2019-04-03
Update the module text box for documentation links

lfnothias:
Hi Tomas,

I am just updating the module description with links to the documentation and tutorial.
Also, @kaibioinfo is in the process of adding a new MS/MS merging feature for that module.


tomas-pluskal:
Awesome, thanks!

PR #561, opened 2019-04-10
Bugfix for Spectral Deconvolution/Multivariate Curve Resolution that we discussed in #553

asmirn1:
This updated spectral deconvolution uses method `Feature.getParentChromatogramRowID()` as we discussed in #553.


tomas-pluskal:
Thanks, Alex!
I think now would be a good time to prepare that sample dataset we've been talking about :)

asmirn1:
@tomas-pluskal Yes, I'm working on updating ADAP User Manual now. I should be able to finish it and add a sample dataset in a few days, maybe a week.

Sorry, it takes so long. I'm doing my best to finish it ASAP.

PR #562, opened 2019-04-12
Changed to new GNPS submission address, added password and username

robinschmid:
Hey Tomas, Hey Ming,

I have added the password and username option to the direct GNPS submission and moved the address to

http://dorresteinappshub.ucsd.edu:5050/uploadanalyzefeaturenetworking

@mwang87 , is this the final address? Or are you planning to change the address again?

The password is not encrypted and just send as a String. What is the best way to handle this?

Cheers
Robin


tomas-pluskal:
I'd say sending a plain-text password over HTTP is not ideal. Does the server support HTTPS?

robinschmid:
I have talked to ming. Their plan is to move the server to a final location with https. 
However, this might take a while. 
As this parameter is optional for GNPS we could just mention it in the help file  the description and maybe in the parameter name - what fo you think?

tomas-pluskal:
Sounds good!

Tomas

robinschmid:
I have changed the parameter name, added some comments in the help file and this should be enough to notify everyone that the password is sent without encryption.
This should be fine to merge.

tomas-pluskal:
Awesome, thanks!

PR #563, opened 2019-04-12
Import GNPS library matching results

robinschmid:
Hey Tomas,

this is a small module that imports all GNPS library matching results from a graphml file that is provided by the GNPS results export. In another branch, that I still need to make ready for merging, the whole GNPS network is imported to MZmine and visualised.

They are only added as identities - but there is much more interesting information in the GNPSResultsIdentity (e.g., the library entry link). Can you think of a good way to enable the user to open the hidden information that also leads to the GNPS webpage)
image

Best
Robin

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -107,8 +107,27 @@
 			<name>MZmine local libraries</name>
 			<url>file://${project.basedir}/src/main/lib</url>
 		</repository>
+		<!--  for graphstream -->
+    <repository>
+        <id>jitpack.io</id>
+        <url>https://jitpack.io</url>
+    </repository>
 	</repositories>
+	
 	<dependencies>
+		<!-- https://mvnrepository.com/artifact/org.graphstream/gs-core 
+		Networks visualisation-->
+		<dependency>
+	    	<groupId>com.github.graphstream</groupId>
+	    	<artifactId>gs-core</artifactId>
+	    	<version>2.0-alpha</version>
+		</dependency>
+		<dependency>
+	    	<groupId>com.github.graphstream</groupId>
+    		<artifactId>gs-ui-swing</artifactId>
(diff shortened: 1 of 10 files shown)


photocyte:
Interesting addition!  Wouldn't this be better as description rather than identity?  A spectral match isn't really an identity, per se

Annexhc:
Hey Robin, nice addition!
Well, I think you can call it an identity. It is up to the user if an identity is valid or not. Anyway I like the idea to use the description or comment column as well. I would suggest to split the annotation. E.g. Identity: Myristoleic acid 
Comment: GNPS spectral match, NIST14 (M+H-H2O)

tomas-pluskal:
I agree we can call it an identity, because the concept of identity is very loosely defined in MZmine (e.g., adducts are also identities etc.). 

As for opening the identity-specific URL, you can add a menu item to https://github.com/mzmine/mzmine2/blob/master/src/main/java/net/sf/mzmine/modules/visualization/peaklisttable/PeakListTablePopupMenu.java like "Open compound ID URL". How about that?

In the long-term, I agree we need to improve the data model to make it more flexible, because different MZmine modules want add different types of information to the peak list table.

robinschmid:
Everyone should always use multiple orthogonal methods to really ensure an identification - in MZmine, identities are everything. Small parts that contribute to the whole identification. 

I don't thnk we should use the comment part for this. Comments are a great and fast way for the user to enter more information. The identity is more complex and flexible. 

The last commit adds a menu item to the peaklist popup menu: 
identities/Open coumpound ID URL

To support any kind of identification method. the menu is enabled if the preferred compound ID has a URL and will open the browser.

tomas-pluskal:
Looks good to me!

PR #568, opened 2019-04-30
Merge MS/MS when exporting for GNPS or SIRIUS

kaibioinfo:
Hi,
I start implementing a merge functionality for MS/MS spectra. The current implementation contains three different merge modes:
- no merge (export all scans)
- merge only scans within the same raw data file
- merge scans across all raw data files (belonging to the same feature row in an aligned feature table)

Furthermore, merging involves the following steps:
- start with the MS/MS scan which has the best precursor (= low chimeric intensity, high precursor intensity, high MSMS TIC)
- iteratively merge surrounding MS/MS IF there cosine similarity is above a user given threshold (this should help in cases you have chimeric contaminations in some of your MSMS)
- several cleaning, outlier removal steps, merging of intensities and masses

The pull request affects the GNPSExport as well as the SIRIUS export

### DIFF
--- a/src/main/java/net/sf/mzmine/modules/peaklistmethods/io/siriusexport/MergeUtils.java
+++ b/src/main/java/net/sf/mzmine/modules/peaklistmethods/io/siriusexport/MergeUtils.java
@@ -1,0 +1,739 @@
+package net.sf.mzmine.modules.peaklistmethods.io.siriusexport;
+
+import com.google.common.collect.Range;
+import net.sf.mzmine.datamodel.DataPoint;
+import net.sf.mzmine.datamodel.MassList;
+import net.sf.mzmine.datamodel.RawDataFile;
+import net.sf.mzmine.datamodel.Scan;
+import net.sf.mzmine.datamodel.impl.SimpleDataPoint;
+import org.apache.commons.math3.special.Erf;
+import org.slf4j.LoggerFactory;
+
+import java.util.*;
+import java.util.stream.Collectors;
+
+/*
+    Merge several MS/MS scans into one
+ */
+public class MergeUtils {
+
+    /**
+     * Peaks below this threshold will not be counted as chimerics. The reasoning is that peaks with low intensity
+     * won't contaminate the MS/MS scans anyways.
+     */
+    public static final double CHIMERIC_INTENSITY_THRESHOLD = 0.1;
+    /**
+     * The masslist to use. If null, always pick all peaks
+     */
+    protected String masslist;
+
+    /**
+     * expected average mass deviation of fragment peaks
+     */
+    protected double expectedPPM;
+
+    /**
+     * width and offset (from precursor m/z) of the isolation window.
+     * Note that the window is centered at the precursor m/z
+     */
+    protected Range<Double> isolationWindow;
+
+    /**
+     * Only merge spectra if their cosine is above this threshold.
+     * Important: If your spectra come from different collision energies, you might want to set this threshold very low!
+     */
+    protected double cosineThreshold;
+
+    /**
+     * If true, the fragment m/z is calculated as weighted average over the merged fragment peaks
+     * If false, the fragment m/z is the m/z of the most intensive fragment peak.
+     *
+     * I guess first option may sometimes have more accurate m/z, while latter one is more safe and will rarely fail
+     */
+    protected boolean mergeMasses;
+
+    /**
+     * Only merge spectra if their MS1 precursor intensity is above this threshold (relative to most intensive precursor
+     * peak).
+     */
+    protected double minPrecursorIntensityRelativeToBestPeak;
+
+    /**
+     * Remove spectra which chimeric contamination is above this threshold
+     */
+    protected double maximalChimericIntensity;
+
+
+    /**
+     * do not merge if the chimeric contamination is larger than X times the contamination of the best peak
+     */
+    protected double maximalChimericContaminationRelativeToBestPeak;
+
+    /**
+     * do not merge if the chimeric contamination is larger than X times the absolute contamination of the best peak
+     */
+    protected double maximalAbsoluteChimericContaminationRelativeToBestPeak;
+
+    public MergeUtils() {
+        this.expectedPPM = 10;
+        this.isolationWindow = Range.closed(-1d,1d);
+        this.cosineThreshold = 0.8;
+        this.mergeMasses = true;
+        this.masslist = null;
+        this.minPrecursorIntensityRelativeToBestPeak = 0.25;
+        this.maximalChimericIntensity = 0.66;
+        this.maximalChimericContaminationRelativeToBestPeak = 1.33;
+        this.maximalAbsoluteChimericContaminationRelativeToBestPeak = 1.33;
+    }
+
+    public String getMasslist() {
+        return masslist;
+    }
+
+    public void setMasslist(String masslist) {
+        this.masslist = masslist;
+    }
+
+    public double getExpectedPPM() {
+        return expectedPPM;
+    }
+
+    public void setExpectedPPM(double expectedPPM) {
+        this.expectedPPM = expectedPPM;
+    }
+
+    public Range<Double> getIsolationWindow() {
+        return isolationWindow;
+    }
+
+    public void setIsolationWindow(Range<Double> isolationWindow) {
+        this.isolationWindow = isolationWindow;
+    }
+
+    public double getCosineThreshold() {
+        return cosineThreshold;
+    }
+
+    public void setCosineThreshold(double cosineThreshold) {
+        this.cosineThreshold = cosineThreshold;
+    }
+
+    public boolean isMergeMasses() {
+        return mergeMasses;
+    }
+
+    public void setMergeMasses(boolean mergeMasses) {
+        this.mergeMasses = mergeMasses;
+    }
+
+    protected double getTIC(DataPoint[] scan, double parentPeakThreshold) {
+        double TIC = 0d;
+        for (DataPoint p : scan) {
+            if (p.getMZ() > parentPeakThreshold)
+                break;
+            TIC += p.getIntensity();
+        }
+        return TIC;
+    }
+
+
+    //////////////////////////////////////////////////////////////////
+
+
+    /**
+     * Merge a list of scans from different origins into one spectrum. Each peak in the merged spectrum remembers its original
+     * peaks.
+     * @param name just for debugging purposes: some name string to identify the origin of the fragment scan
+     * @param parentPeak the m/z of the precursor ion
+     * @param scans a list of scans which should be merged
+     * @return merged spectrum
+     */
+    public MergedSpectrum mergeScansFromDifferentOrigins(String name, double parentPeak, List<FragmentScan> scans) {
+        if (scans.isEmpty()) return MergedSpectrum.empty();
+        // first split of list by origins
+        final HashMap<String, List<FragmentScan>> byName = new HashMap<>();
+        for (FragmentScan s : scans) {
+            byName.computeIfAbsent(s.origin.getName(), (x) -> new ArrayList<FragmentScan>()).add(s);
+        }
+        final List<List<FragmentScan>> allFs = new ArrayList<>(byName.values().stream().filter(x->!x.isEmpty()).collect(Collectors.toList()));
+        if (allFs.size()==1) {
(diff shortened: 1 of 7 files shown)


src/main/java/net/sf/mzmine/modules/peaklistmethods/io/siriusexport/MergeUtils.java
```diff
@@ -0,0 +1,1156 @@
...
+            LoggerFactory.getLogger(MergeUtils.class).warn(name + ": Only " + numberOfMerges + " of " + mzOrderedCopy.size() + " spectra are merged. All other have a too low cosine similarity. Cosines between most intensive scan and other scans are: " + Arrays.toString(cosines));
+        }
+        MergedSpectrum M = new MergedSpectrum(mergedSpectrum, new RawDataFile[]{scans.get(0).origin}, scanNumbers.stream().mapToInt(x -> x).toArray(), polarityType, charge,numberOfScansRemovedDueToLowQuality, numberOfScansRemovedDueToLowCosine);
+        if (cutoffOutliers) M = cutOffOutliers(M);
+        if (!innerMerge) M = filterOutRarePeaks(M);
+        return M;
+    }
+
+
+    public List<FragmentScan> extractFragmentScansFor(PeakListRow row, Feature f) {
```
tomas-pluskal: There is a function ```f.getAllMS2FragmentScanNumbers()``` to extract all fragment scans - what else is this method doing? Perhaps some JavaDoc comments for the methods would be useful.

tomas-pluskal:
Thanks Kai, overall it looks good but the MergeUtils class seems a bit bloated with many inner classes and few JavaDoc comments... 

I don't like too much introducing so many new classes that will require maintenance but cannot be reused by other modules. For example, ```MergeDataPoint.cutoffOutliers()``` makes little sense to me as a method of a data point - shouldn't that go into a common utility class (e.g., ScanUtils) with a signature like ```public static DataPoint[] cutoffSpectrumOutliers(DataPoint spectrum[])```?

kaibioinfo:
> I don't like too much introducing so many new classes that will require maintenance but cannot be reused by other modules.

As long as they are inner classes, they are not part of the API (and, thus, do not require maintenance). I see this rather as advantage, because the whole stuff is just a quick solution and I would really appreciate if somebody else finds some time in future to do it right.
However, I think a few inner classes are public. I can change their visibility to package local.

>  For example, `MergeDataPoint.cutoffOutliers()` makes little sense to me as a method of a data point - shouldn't that go into a common utility class (e.g., ScanUtils) with a signature like `public static DataPoint[] cutoffSpectrumOutliers(DataPoint spectrum[])`?

I think these stuff is quite specific and should not be put into a common utility class. However, the cosine calculation might be helpful for other modules.

I agree that the code lacks documentation. I will add some additional comments. For me it is rather important to have a working and robust way to export MS/MS spectra. I'm fine with extracting such a feature into an own module.

src/main/java/net/sf/mzmine/modules/peaklistmethods/io/siriusexport/MergeUtils.java
kaibioinfo: f.getAllMS2FragmentScanNumbers() never did work for me. Maybe I have to reanalyze a project from raw data to see any effect?
Otherwise, this method will not only collect MSMS scans but also store the MS1 scans that precede the MSMS.

tomas-pluskal: It's a feature that was added recently, so if you're working with an older project file, the MS2 mapping won't be there. Reprocessing the raw data is one way to go, but Robin also made a separate module to just assign the MS2 scans to features: https://www.youtube.com/watch?v=EL5pmFvpTFE&feature=youtu.be

tomas-pluskal: If retrieving the MS1 scan before the MS2 is something that you need, perhaps we should add such function to the data model. What is the reason behind?

tomas-pluskal:
We also have the SIRIUS search module that performs direct searches using your public API. Currently that module does not perform any merging, and it would be nice to have the behavior consistent.

I actually have somewhat mixed feelings about merging the MS2 spectra. What is the advantage, in comparison to giving the original spectra to SIRIUS and letting it handle them on its own (which might include merging of course)?

src/main/java/net/sf/mzmine/modules/peaklistmethods/io/siriusexport/MergeUtils.java
kaibioinfo: Okay, I will check that.

So far there is only the possibility to get the "best" MSMS or all MSMS. What we need is a function that retrieves all "qualitative" MSMS spectra. The quality control could then be outsourced in an extra module.

The quality of a MSMS spectrum is affected by:
- intensity of the precursor ion in MS1
- how many high intensive peaks are within the isolation window if the precursor ion (chimeric spectra)
- total ion count of the MSMS spectrum
- how similar is the MSMS spectrum to other MSMS spectra of the same feature (cosine similarity)

kaibioinfo:
> We also have the SIRIUS search module that performs direct searches using your public API. Currently that module does not perform any merging, and it would be nice to have the behavior consistent.
> 
> I actually have somewhat mixed feelings about merging the MS2 spectra. What is the advantage, in comparison to giving the original spectra to SIRIUS and letting it handle them on its own (which might include merging of course)?

That is possible, too. But still we would need some quality check to only export MS2 which has some decent quality. Usually, most MS2 spectra belonging to a feature are garbage (they are shot during the begin or end of the peak shape, they intervene with other coeluting peaks with similar mass and so on). We have to exclude these spectra, otherwise the merged spectra will have a much lower quality than just the most intensive MS2 scan.

src/main/java/net/sf/mzmine/modules/peaklistmethods/io/siriusexport/MergeUtils.java
tomas-pluskal: I like that approach! We can use the existing ```f.getAllMS2FragmentScanNumbers()``` function, and then have a separate module that can evaluate the quality of the MS2 spectra and for example provide a Comparator to sort them. Then the export module can choose to select e.g. the 5 highest quality spectra and export those.

Does the precursor ion intensity really matter? I think what matters is the TIC of the MSMS, but whether that TIC was produced from a high-intense precursor or not is irrelevant. Remember that the precursor scan might be quite distant (in terms of retention time) from the MSMS scan, or it might not even exist (targeted MS2) or there might be weird situations like MSe, all ion fragmentation, etc.

robinschmid:
Hey Kai,
Hey Tomas,

please see my PR #569, in which I highlight some classes which might be of help or redundant in the end.

Some classes for different similarity measures and MS2 scan utility methods and sorting

The sorting and other methods always work on DataPoint arrays (scan -> masslist -> DataPoint[]) as we always need masslists for each scan (to avoid problems with profile mode scans).

Cheers
Robin

tomas-pluskal:
Hi Kai,
If you need this function ASAP, I can merge this PR for now and we can rewrite the code later using the existing utility classes that Robin highlighted. What's your suggestion?

tomas-pluskal:
Hi @kaibioinfo @robinschmid 
I added some additional functions to the ScanUtils class:

* selectBestMS2Scans() to select top N best fragmentation scans
https://github.com/mzmine/mzmine2/blob/master/src/main/java/net/sf/mzmine/util/scans/ScanUtils.java#L764
* findPrecursorScan() to find the precursor scan corresponding to MS2
https://github.com/mzmine/mzmine2/blob/master/src/main/java/net/sf/mzmine/util/scans/ScanUtils.java#L742

The best MS2 selection algorithm currently only sorts the MS2 scans by TIC. We can add additional logic based on what you suggested. I like the idea of checking for contaminating peaks in the MS1 isolation window, but we'll need to add the concept of isolation window to the MZmine data model, because currently isolation windows are not parsed from the input files.

Regarding merging of scans - I don't see the point and don't really like it. If you have a good argument why merging should be performed, please explain to me. Otherwise, I would like to remove the merging feature and simply export N best quality MS/MS spectra, where N can be an arbitrary choice (e.g., 10?). Please let me know your opinion, though.

lfnothias:
Hi @tomas-pluskal,
I am working with @kaibioinfo and @robinschmid on the development of the GNPSExport. 
It has been shown that the merging of MS/MS spectra improves the subsequent spectral annotation (library match, better S/N for fragment ions, etc). There are many papers in proteomics in metabolomics showing that (http://theses.gla.ac.uk/7928/ ; 
https://onlinelibrary.wiley.com/doi/full/10.1002/pmic.20170045 ; also MS-Cluster used for GNPS https://europepmc.org/articles/PMC2533155.). The limitation is that it must be done with high quality spectra, or the merged spectra can get worst obviously. That's why Kai is looking for the MS1 first, to ensure that the spectrum is not chimeric (contaminated by isobaric ions). This is done with MS-Purity (https://pubs.acs.org/doi/abs/10.1021/acs.analchem.6b04358) and also in OpenMS (PurityScore function).
All of this is motivated by the fact that GNPS molecular networking accepts one spectrum per "feature". The current GNPSExport module exports the most intense scan which can be contaminated or of low intensity. The present approach allows to make sure we use clean spectra and improve the S/N for fragments. Yet users have the possibility to select the option of choice.

Our preliminary results are showing that many most intense spectra are chimeric + better spectral matches for the merged spectra. But we need do a more systematic benchmarking and it will take time.

I hope that's detailed enough. That's really exciting to see all the new features in the last MZmine release ! I am big fan of the spectral matching modules and the spectral library entry interface by @robinschmid

tomas-pluskal:
Hi @lfnothias !
Thanks for the info. As you said, the motivation behind merging spectra is the limitation on the GNPS side (one spectrum per feature). However, that is not the case for SIRIUS, which can accept batches of spectra for the same feature.

Currently the SIRIUS export module performs merging, but the GNPS module doesn't... so we should be doing it the opposite way, shouldn't we?

lfnothias:
Indeed 👍 
The merging of high quality spectra is targeting ultimately the GNPSExport, and not Sirius since multiple spectra can be inputted per feature. I am not hundred percent sure, but I think there is some merging happening in Sirius eventually anyway (@kaibioinfo). Also, for very large projects with hundred of files, the merged spectrum might be interesting to use with the Sirius export, since files can reach several gigabytes in some case.
I believe Kai started implementing the merge function into the Sirius export as it is easier for him to debug and evaluate.

tomas-pluskal:
@kaibioinfo Is there any preference for SIRIUS regarding the type of spectra is should receive as input (merged/unmerged)?

kaibioinfo:
> I believe Kai started implementing the merge function into the Sirius export as it is easier for him to debug and evaluate.
right ;)

@tomas-pluskal the problem is that in SIRIUS we no longer know about the quality of the input spectra. That's why we merge very conservatively. If you pick the MS/MS yourself and ensure that they are good quality, you can do better merging (e.g. averaging their m/z values, noise peak removal, and so on).

tomas-pluskal:
@kaibioinfo Okay so let's make a separate module that will do this properly, will clean API and well-defined parameters. This module can be then used for SIRIUS export, GNPS export, as well as for the direct SIRIUS search.

The correct way to do this is to include an OptionalModuleParameter (https://github.com/mzmine/mzmine2/blob/master/src/main/java/net/sf/mzmine/parameters/parametertypes/submodules/OptionalModuleParameter.java) into these three modules, with an embedded parameter set of the merging module.

By the way, the mass lists in MZmine should already be noise-filtered, so I would prefer not to introduce another noise filtering step here.

kaibioinfo:
Alright.

Regarding noise filter: The (optional) filter step is removing peaks which do occur in less than X % of the merged spectra. This cannot be done in the Masslist-step (as it depends on the merging). However, I use the masslists in my code.

tomas-pluskal:
Ok. Could you please move the code into a new package, e.g. 
```net.sf.mzmine.modules.tools.msmsspectramerge```

and create a class for it

```public class MsMsSpectraMerge implements MZmineModule```

with a corresponding ParameterSet?

kaibioinfo:
I'm on it.

kaibioinfo:
Okay, I'm cleaned up the code. I will make a new push request, as I'm not sure how to clean up this branch propery.

Issue #570, opened 2019-05-02
ADAPChromatogramBuilderTask fails


glerny:
### Setup

I am reporting a problem with MZmine 2 version 2.39 and before 

My operating system is  64-bit.

Below I pasted the exact output of the following commands:
```
java -version

[12:14:18|INFO|ADAPChromatogramBuilderTask]: Started chromatogram builder on a-SampleA_1
[12:14:19|SEVERE|WorkerThread]: Unhandled exception java.lang.IllegalStateException: Incorrect range [NaN, NaN] for m/z NaN while processing task Detecting chromatograms in a-SampleA_1(java.lang.IllegalStateException: Incorrect range [NaN, NaN] for m/z NaN@net.sf.mzmine.modules.masslistmethods.ADAPchromatogrambuilder.ADAPChromatogramBuilderTask.run(ADAPChromatogramBuilderTask.java:332))
java.lang.IllegalStateException: Incorrect range [NaN, NaN] for m/z NaN
        at net.sf.mzmine.modules.masslistmethods.ADAPchromatogrambuilder.ADAPChromatogramBuilderTask.run(ADAPChromatogramBuilderTask.java:332)
        at net.sf.mzmine.taskcontrol.impl.WorkerThread.run(WorkerThread.java:58)

``

robinschmid:
Hey

please give us all the parameter values that you have entered for mass detection and the ADAP chromatogram builder.

It seems like your input for m/z tolerance is wrong. NaN means "not a number".

Cheers
Robin

glerny:
Hi Robin,

You can find the print screen of the conditions attached. I believe it is more a problem with the mzML files (converted from TIF to mzML with MSConvert) before or after the MassDetection step (exact mass with a threshold at 1,000) as the same parameters work with other files. I assume the NaN values result from one of those two steps and this case is not taken into account by the ADAP algorithms.  I just test it using a threshold at 10,000 for the MassDetection (exact mass with a threshold at 10,000) and in this case, the ADAP algorithms work, so I definitely believe that during the centroidisation some errors occur that generate NaN values that bugged the ADAP algorithms (the chromatogram builder work through).  If you want to test the mzML files are here:  https://drive.google.com/drive/folders/1dP_fcDE1U7bDrVrMVajUc1NkRz7SVDjr?usp=sharing


![ADAP](https://user-images.githubusercontent.com/11735178/57076025-743ade80-6ce0-11e9-9516-8f638e179b19.jpg)

Best

[...]

PR #572, opened 2019-05-03
Bugfix for #570

asmirn1:
Hi Tomas,

@glerny is right in his description of this bug:
> I believe it is more a problem with the mzML files (converted from TIF to mzML with MSConvert) before or after the MassDetection step (exact mass with a threshold at 1,000) as the same parameters work with other files. I assume the NaN values result from one of those two steps and this case is not taken into account by the ADAP algorithms. I just test it using a threshold at 10,000 for the MassDetection (exact mass with a threshold at 10,000) and in this case, the ADAP algorithms work, so I definitely believe that during the centroidisation some errors occur that generate NaN values that bugged the ADAP algorithms (the chromatogram builder work through).

I have added a check for NaN values in ADAPChromatogramBuilder.

### DIFF
--- a/src/main/java/net/sf/mzmine/modules/masslistmethods/ADAPchromatogrambuilder/ADAPChromatogramBuilderTask.java
+++ b/src/main/java/net/sf/mzmine/modules/masslistmethods/ADAPchromatogrambuilder/ADAPChromatogramBuilderTask.java
@@ -258,7 +258,7 @@ public class ADAPChromatogramBuilderTask extends AbstractTask {
         return;
       }
 
-      if (mzPeak == null) {
+      if (mzPeak == null || Double.isNaN(mzPeak.getMZ()) || Double.isNaN(mzPeak.getIntensity())) {
         // System.out.println("null Peak");
         continue;
       }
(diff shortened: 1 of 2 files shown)


tomas-pluskal:
Thanks, Alex, but where do the NaN values come from? They should not be there.

asmirn1:
It's probably happens during mass detection. I can check it.

tomas-pluskal:
Please take a look, thanks!

asmirn1:
Function `calculateExactMass` in ExactMassDetector.java produces these NaN values. I'll see if I can fix it.

asmirn1:
So the function `calculateExactMass` in ExactMassDetector.java would give NaN values when interpolating between two data points. In case of the zero slope between two data point, the resulting m/z value would be NaN. 

I have modified the code so that when a slope is zero, the interpolated m/z value is the average of two m/z values. Hopefully, this makes sense.

tomas-pluskal:
That's perfect, thanks Alex!

asmirn1:
Tomas,

I'm not sure why my code doesn't pass the check. Is it something with Java versions?

tomas-pluskal:
Don't worry about that, there is some ongoing issue with the mvnsearch.com repository, so the test builds just time out.

PR #573, opened 2019-05-06
Feature/export csv/peak name

asmirn1:
Tomas,

Could we add an option to export a peak name to a .csv file? It can serve as a peak's identifier after alignment.

Alex.

### DIFF
--- a/src/main/java/net/sf/mzmine/modules/peaklistmethods/io/csvexport/CSVExportTask.java
+++ b/src/main/java/net/sf/mzmine/modules/peaklistmethods/io/csvexport/CSVExportTask.java
@@ -332,6 +332,9 @@ public class CSVExportTask extends AbstractTask {
               case PEAK_STATUS:
                 line.append(peak.getFeatureStatus() + fieldSeparator);
                 break;
+              case PEAK_NAME:
+                line.append(peak.toString() + fieldSeparator);
+                break;
               case PEAK_MZ:
                 line.append(peak.getMZ() + fieldSeparator);
                 break;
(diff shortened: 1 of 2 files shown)


tomas-pluskal:
Hi Alex,
Sure, we can do that, but wouldn't it make more sense to add the "Raw data file name" as an option? The peak name looks a bit weird in the CSV format, because it is just plain text with poorly defined structure, but it contains redundant information with the other fields. What do you plan to use this "identifier" for?

asmirn1:
I'm using it to match features before and after alignment. 

For instance, I export all unaligned GC spectra to .msp files, export alignment results to a .csv file, and then I want to match spectra/features from .msp and .csv files.

Although peak name is redundant, I think it would be nice to have it because the peak name in .csv files matches the feature name in the corresponding .msp files. Otherwise, I would have to match the raw data file name, retention time (plus/minus tolerance), and m/z value (plus/minus tolerance).

tomas-pluskal:
But why do you need to match features before and after alignment?

Using the generated String (peak name) as an ID for matching the peaks seems non-ideal to me... If the user performs any adjustments to the peaks (e.g. using the manual peak picker), your IDs are going to be messed up.

asmirn1:
Sure, the peak name is not an ideal identifier. But in the current MZmine, there is no any peak/feature identifier. Combination of filename, m/z, and retention time is good for manual matching but not for automation.

Anyway, I just find it convenient in my work. I don't know if it will be useful for anyone else.

tomas-pluskal:
Ok, I can merge this since it doesn't break any existing functions. But if this is something you want to do long-term, I would recommend to find a more "proper" solution, e.g. add a new method to Feature like getUniqueFeatureID(), which would allow the modules to track the same feature. The ID would not change after alignment or adjustments to the peak with manual peak picker.

asmirn1:
Thank you, Tomas. I'll think about getUniqueFeatureID().

PR #575, opened 2019-05-09
Spectra matching

Annexhc:
Hello Tomas,
based on @robinschmid s amazing MS/MS similarity algorithms I have added a spectra matching module to the spectra visualizer. This enables a fast compound identification for GC-HRMS data. Furthermore, I have added the JCAMP-DX database format. The spectra matching should work with jdx, jason, and msp. 
The result frame displays database information depending on the database input. 
I hope you like it.
Future plans include MS1 Spectra matching also for feature lists. I will do this in an other pull request.
best wishes,
Ansgar

Spectra_Matching__MS1_GC
Spectra_Matching__MS1_GC


Annexhc:
Ups, I have added two times the same gif. This gif shows the possibility for MS/MS scans using a json database, which was generated with @robinschmid  MS/MS export library.

Spectra_Matching__MSMS

tomas-pluskal:
Thanks Ansgar, that looks great!

For the SpectralDBParser and related classes, wouldn't it make sense to put them into a separate package (i.e., not under the simplespectra visualizer)? These parsers could be useful for other modules.

src/main/java/net/sf/mzmine/modules/visualization/spectra/simplespectra/spectraidentification/spectraldatabase/parser/NistMspParser.java
```diff
@@ -0,0 +1,128 @@
...
+import net.sf.mzmine.datamodel.impl.SimpleDataPoint;
+import net.sf.mzmine.main.MZmineCore;
+import net.sf.mzmine.modules.peaklistmethods.identification.spectraldbsearch.dbentry.DBEntryField;
+import net.sf.mzmine.modules.peaklistmethods.identification.spectraldbsearch.dbentry.SpectralDBEntry;
+import net.sf.mzmine.modules.visualization.spectra.simplespectra.SpectraPlot;
+import net.sf.mzmine.modules.visualization.spectra.simplespectra.spectraidentification.spectraldatabase.SpectralMatchTask;
+import net.sf.mzmine.parameters.ParameterSet;
+import net.sf.mzmine.taskcontrol.AbstractTask;
+
+public class NistMspParser implements SpectralDBParser {
```
robinschmid: Isnt this just a copy of the original one?
You could use the other one correct?

robinschmid:
All of the parsers (except for jdx) are a copy of the original ones? You could move your parser into the original directory

robinschmid:
Great addition btw

src/main/java/net/sf/mzmine/modules/visualization/spectra/simplespectra/spectraidentification/spectraldatabase/parser/NistMspParser.java
Annexhc: Hey Robin,
no the parsers are different. They dont have a boolean for precursorMZ. But I agree with Tomas that we can place the parsers somewhere else, including your original ones.

robinschmid:
We should just change the precursor mz to be optional - would that be possible?

Annexhc:
I am not sure - Does it make sense to check for the precursor already in the DB parser? That makes only sense to me if the parser is specifically designed for MS/MS databases. If we want to reduce the number of parsers we should parse everything and do check ups in the modules.
I think it should be okay to have two parsers, one for MS1 level one for MS/MS spectra. What do you think?

Annexhc:
Following Tomas suggestion. We could make a package in utils (e.g. databases) and add the parser packages there. One for MS1, one for MSMS

robinschmid:
The data formats/libary entry formats are the same. Therefore, we should only have one parser per file format.

Annexhc:
Okay, so what are your suggestions regarding my idea to remove the boolean from the parser?

robinschmid:
> Okay, so what are your suggestions regarding my idea to remove the boolean from the parser?

            if (entry.getPrecursorMZ() != null) {
              list.add(entry);
              // progress
              if (list.size() % 1000 == 0) {
                // start new task for every 1000 entries
                logger.info("Imported " + list.size() + " library entries");
                SpectralMatchTask task =
                    new SpectralMatchTask(peakList, parameters, tasks.size() * 1000 + 1, list);
                MZmineCore.getTaskController().addTask(task);
                tasks.add(task);
                // new list
                list = new ArrayList<>();
              }
            }

I think you are concerned about this part?
We could just leave the checking to the module which is going to use the library entries. Or define a boolean parameter for this method to control if the precursor mz is mandatory.

Another thing:
The reason why a SpectralMatchTask is generated for every 1000 entries is to allow very large databases. To allow different tasks (MS/MS, your GC spectral match....) we could implement something like a listener/Consumer<List<SpectralDBEntry>>.

Annexhc:
Yes, that is the part. I really like to move the checking to the module which is going to use the library entries. I will see what I can do. Hopefully I wont break your modules ;-)

tomas-pluskal:
> Ups, I have added two times the same gif. This gif shows the possibility for MS/MS scans using a json database, which was generated with @robinschmid MS/MS export library.
> 
> Spectra_Matching__MSMS

I noticed in this GIF there is no mass list used. Are you matching against the raw data points of the spectrum? That would cause problems with profile mode spectra. You should always use a mass list (add the MassListParameter to the dialog).

Annexhc:
Hey Tomas,
no I match against a mass list. The mass list is generated automatically. The algorithm checks for profile or centroid and then uses the exact mass algorithm for profile and the centroid one for centroid. I know, that is not perfect if we have e.g. TOF data. I can a the mass list parameter. Makes it easier for me in the end.

tomas-pluskal:
If you do that kind of automation, it defeats the purpose of having the mass lists stored :) Better to use the parameter.

Annexhc:
True, I will add it

Annexhc:
Hey Tomas,
I think I have addressed all ideas and concerns. Robin and I have made some changes regarding the DB parsers. We are now using the same parser for MS1 and MSMS scans. Parsers are moved to utils. All further logic, such as precursor m/z check, are now moved to the modules. I made some changes to the result window as well.

src/main/java/net/sf/mzmine/modules/peaklistmethods/identification/spectraldbsearch/LocalSpectralDBSearchTask.java
```diff
@@ -110,37 +108,36 @@ public String getTaskDescription() {
...
-        throw new MSDKRuntimeException(
-            "DB file was empty - or error while parsing " + dataBaseFile);
       }
-    } catch (Exception e) {
-      logger.log(Level.SEVERE, "Could not read file " + dataBaseFile, e);
-      setStatus(TaskStatus.ERROR);
-      setErrorMessage(e.toString());
-      throw new MSDKRuntimeException(e);
+      try {
+        Thread.sleep(100);
```
tomas-pluskal: What is the purpose of this Thread.sleep()?
It looks suspicious. Perhaps add some comments to explain what you're doing.

src/main/java/net/sf/mzmine/modules/visualization/spectra/simplespectra/spectraidentification/spectraldatabase/SpectraIdentificationSpectralDatabaseTask.java
```diff
@@ -0,0 +1,192 @@
...
+          try {
+            Thread.sleep(100);
+          } catch (Exception e) {
+            cancel();
+          }
+        }
+      } else {
+        setStatus(TaskStatus.ERROR);
+        setErrorMessage("DB file was empty - or error while parsing " + dataBaseFile);
+        throw new MSDKRuntimeException(
```
tomas-pluskal: Why throw new exception here? After setting the status to ERROR, you should just ```return```.

src/main/java/net/sf/mzmine/modules/visualization/spectra/simplespectra/spectraidentification/spectraldatabase/SpectralMatchTask.java
```diff
@@ -0,0 +1,242 @@
...
+
+      // check if no match was found
+      if (count == 0) {
+        logger.log(Level.WARNING, "No data base matches found");
+        setErrorMessage("No data base matches found. Spectral data base matching failed");
+        list = null;
+        return;
+      }
+
+    } catch (Exception e) {
```
tomas-pluskal: Do not swallow exceptions - at least add ```e.printStackTrace()```.
Also, I think it is generally confusing to use long try/catch blocks. You should just take those parts of the code that may throw an exception, and surround those with try/catch.

src/main/java/net/sf/mzmine/modules/peaklistmethods/identification/spectraldbsearch/LocalSpectralDBSearchTask.java
robinschmid: This loop is just waiting for all sub tasks to finish - and then it prints out the total number of library hits and finishes the main task.
Thread.sleep(100); is just to not waste too much time in this thread. 
Any better idea?

src/main/java/net/sf/mzmine/modules/visualization/spectra/simplespectra/spectraidentification/spectraldatabase/SpectralMatchTask.java
robinschmid: Maybe it is better to put this try-catch into the loop.
Otherwise the result of the task should be error

src/main/java/net/sf/mzmine/modules/visualization/spectra/simplespectra/spectraidentification/spectraldatabase/SpectraIdentificationSpectralDatabaseTask.java
Annexhc: approved

src/main/java/net/sf/mzmine/modules/visualization/spectra/simplespectra/spectraidentification/spectraldatabase/SpectralMatchTask.java
Annexhc: Yeah, that was bad practice. I will add an exception and simplify the code.

Annexhc: I will set it to error

src/main/java/net/sf/mzmine/modules/peaklistmethods/identification/spectraldbsearch/LocalSpectralDBSearchTask.java
tomas-pluskal: It’s fine, maybe just add some comments to explain

Annexhc: will do

tomas-pluskal:
Thanks, looks great. I'll merge this now.

PR #577, opened 2019-05-09
Faster noise peak removal

kaibioinfo:
Hi,

this is not a "serious" push request - as it is just a quick hack to demonstrate how to make noise peak removal faster. To implement it, we need some changes in the core API of MZMine.

I processed some experiment with MZMine and the whole analysis took almost two hours. Most time was just spent on the "mass detection (centroid)" step. After changing some lines of code the whole analysis runs within 5 minutes instead of two hours.

The reason for this is the inefficient implementation of spectra within MZMine. Unfortunately, as this is a core data structure in MZmine, we cannot easily change that. But I saw that MSDK is implementing spectra in a more efficient way (arrays of mzs and intensities), so maybe the problem will be solved with MZMine 3? However, the first step of every analysis is usually some noise removal. Even very low intensity baselines will usually result in a data size reduction of 99% or more. If it is possible to avoid the allocation of DataPoints before noise removal, we can significantly speed up the whole processing workflow.

I would suggest to have so method that allows to read and write the FloatBuffer of spectra directly. Or any other method that allows to access mass spectral data in an efficient way.  

Beside that: We should DEFINITELY change the MassDetector Task such that it does not store all the peaks when saveToCDF is false.

Furthermore, I would recommend to use the Trove4J library instead of ArrayList<Double> to save memory.


tomas-pluskal:
Hi Kai, I totally agree that a lot of optimization can be made in the algorithms in MZmine. I am planning to switch completely to the MSDK data model in the next few months, but for now I am happy to merge this change, as long as it doesn't break any existing functionality.

The PR contains commits from the other PR, including that copy-pasted garbage... Could you make a new PR with just the single commit relevant to this?

lfnothias:
Wow. That's really impressive !  ^^
From our experience, I can confirm that mass detection takes the longest when we do large scale data processing (several hundreds or thousand), or when we open up projects.

kaibioinfo:
Okay, I clean that up and make a new pull request.

PR #578, opened 2019-05-10
Spectra matching

robinschmid:
- moved parsers to util.spectraldb
- Same parsers for different tasks/modules
- Process entries in batches of 1000 library entries
- precursor mz is not mandatory
- jdx import

Hey Ansgar,

add the changes that you have done to the msp parser and other parsers to the new structure and delete your duplicates. 
At this stage, your spectralMatchTask will create a new Window (or error) for every 1000 database entries. You should directly open one results window and use a synchronized method to add and sort all results to the same window.

Cheers,
Robin


tomas-pluskal:
Is this based on #575? Should we close #575?

robinschmid:
Oh sorry Tomas,
I wanted to directly create a pull request for ansgars branch in his repo. 
He still needs to check some things.

Cheers
Robin

PR #580, opened 2019-05-11
Fix slow mass detection

kaibioinfo:
Hi,

can you roll back my last pull request about the faster mass detection?

I found out that the DataPoint structure is not the problem. In fact, when parsing the MzXML file, all spectra are converted into this format anyways. 

The real problem is that you add a new SWING event to the event queue every time you add a new mass list (which is done for every spectrum). So if you have ten thousands of spectra in a file, MZmine is busy for hours just updating the event queue.

I add a new fix which just delays the update until all scans of a raw data file are processed. The advantage of this fix: it does not introduce any strange hack. It has only very small changes in the core API. It is much cleaner than before. And it is super fast - reducing the time for mass detection from hours to seconds.

### DIFF
--- a/src/main/java/net/sf/mzmine/datamodel/Scan.java
+++ b/src/main/java/net/sf/mzmine/datamodel/Scan.java
@@ -18,10 +18,10 @@
 
 package net.sf.mzmine.datamodel;
 
+import com.google.common.collect.Range;
+
 import javax.annotation.Nonnull;
 import javax.annotation.Nullable;
-
-import com.google.common.collect.Range;
 
 /**
  * This class represent one spectrum of a raw data file.
@@ -94,5 +94,12 @@ public interface Scan extends MassSpectrum {
   public void addMassList(@Nonnull MassList massList);
 
   public void removeMassList(@Nonnull MassList massList);
+
+  /**
+    This will add a mass list, but won't trigger an update in the event queue.
+   You have to call RawDataFile#notifyUpdatedMassLists() to trigger the update afterwards
+   * @param massList
+   */
+  public MassList addMassListWithoutNotification(@Nonnull MassList massList);
 
 }
--- a/src/main/java/net/sf/mzmine/datamodel/RawDataFile.java
+++ b/src/main/java/net/sf/mzmine/datamodel/RawDataFile.java
@@ -18,9 +18,11 @@
 
 package net.sf.mzmine.datamodel;
 
+import com.google.common.collect.Range;
+
 import javax.annotation.Nonnull;
 import javax.annotation.Nullable;
-import com.google.common.collect.Range;
+import java.util.List;
 
 public interface RawDataFile {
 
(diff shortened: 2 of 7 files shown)


kaibioinfo:
Forget the rollback, I just removed the code from the last pull request myself. Sorry for messing up your git history ^^°

tomas-pluskal:
Okay, that makes sense.

Just to clarify: after parsing mzXML, the spectra are converted to DataPoint[] array, but when the new Scan is added to the RawDataFileImpl class, it is changed into StorableScan and the DataPoint[] array is stored in a temporary file to save memory. It is not a very smart or efficient design, but that's the way it was written long time ago and nobody really wants to spend time improving it..

Another comment: I have never experienced the mass detector running for hours. I suppose this could have something to do with how many raw data files you are processing in parallel and how many scans are in those files. That might be why it's causing Swing thread bottlenecks for you, while I have not experienced it with my files. There is also a possibility that it could be some weird Swing issue that only appears on your system (like #405).

src/main/java/net/sf/mzmine/datamodel/Scan.java
```diff
@@ -95,4 +95,11 @@
 
   public void removeMassList(@Nonnull MassList massList);
 
+  /**
+    This will add a mass list, but won't trigger an update in the event queue.
+   You have to call RawDataFile#notifyUpdatedMassLists() to trigger the update afterwards
+   * @param massList
+   */
+  public MassList addMassListWithoutNotification(@Nonnull MassList massList);
```
tomas-pluskal: Why does this return MassList instead of void?

tomas-pluskal:
Perhaps you could keep the list of the newly added StorableMassLists inside the RawDataFileImpl class and change the update method to 
````
public void notifyUpdatedMassLists();
````
That would reduce the confusion of passing around Lists of MassLists..

kaibioinfo:
I didn't try it on Windows, but adding a single event to the event dispatcher for every spectrum in a dataset is clearly an overkill. 

I would suggest to let the Scan class never fire any event.

kaibioinfo:
To justify this: also RawDataFile#finishWriting has to be called at the end of all scan modifications. So I think it's fine to do the same with RawDataFile#addMassLists or something similar.

tomas-pluskal:
I am not objecting against your solution :)

My comments were just about the details of the implementation.

kaibioinfo:
Actually, as it is a modification of some core classes (Scan and RawDataFile), I'm fine if you just decide for the API change. But I think it is important to have some way to update all masslists and fire an event change afterwards.

tomas-pluskal:
OK, I'll just merge this and then do some changes myself.

tomas-pluskal:
I moved the GUI updating logic to the RawDataTreeModel class where I think it makes most sense.

https://github.com/mzmine/mzmine2/commit/66374684522fa42409606ff026b70a186885d78a

PR #587, opened 2019-05-16
Fix nio.buffer.clear for jdk>8

robinschmid:
Hey Tomas,

I have fixed this error before but I have added some comments this time. It got changed back in the last PR of Kai. 

When building with jdk>8 and running with JRE8 buffer.clear will throw a NoSuchMethodException. The fix is tested.

The other change is just a renaming of the MS/MS database search.

Cheers
Robin


tomas-pluskal:
Awesome, thanks!

PR #588, opened 2019-05-16
Peaklist spectra matching

Annexhc:
Hi Tomas,
I made some small changes to the MS/MS spectra matching module for peaklists. I have implemented an MS level parameter. Now the module can be used for MS1 scans as well (GC-EI-MS).
@robinschmid pls check if I have cracked your module with this change. Also pls check if I have again changed the rawDataImpl file you have changed earlier. I had a liitle GitHub trouble today :-P


robinschmid:
Seems to be fine. 
I will test it and also add some more changes to this module

tomas-pluskal:
Looks good to me! 
@robinschmid I'll merge this now and you can submit further changes in another PR.

PR #589, opened 2019-05-16
Removed error from msp parser

Annexhc:
Hi Tomas,
I found a bug in the msp DB parser. The parser was handling every data point pair as a DB Entry. Now every is fine again.
best wishes

### DIFF
--- a/src/main/java/net/sf/mzmine/util/spectraldb/parser/NistMspParser.java
+++ b/src/main/java/net/sf/mzmine/util/spectraldb/parser/NistMspParser.java
@@ -91,7 +91,7 @@ public class NistMspParser extends SpectralDBParser {
               }
             }
           }
-          if (isData) {
+          if (l.isEmpty()) {
             // empty row after data
             // add entry and reset
             SpectralDBEntry entry =


tomas-pluskal:
How about adding short comments to the mysterious variables like 
````
int correct = 0;
boolean isData = false;
int sep = -1;
````
Having the variables defined might help to avoid similar problems in the future.

Annexhc:
Yes, I agree! It is especially important when we are working together on the same classes. And you are also right that some of these variables are indeed mysterious, because it seems they are leftovers from the previous data structure....

tomas-pluskal:
@Annexhc @robinschmid 
Should I merge this now, or are you planning to add more comments to the code?

robinschmid:
Hey Ansgar and Tomas,

this bug was introduced in your last PR in commit 6d4aa3c

The correct version was
else if (isData)  versus the current  if(isData)

I am currently working on an mgf parser and I can fix the msp parser. 


Cheers
Robin

Annexhc:
alright Robin, I will close this pull request.

PR #592, opened 2019-05-17
FileNameListParameter to save last files used for a module (batch)

robinschmid:
This was one of my most wanted for a long time now.

This parameter saves the last used files. By double click on a file it can be used from a Consumer<File> or a filename component.

To exemplify, this parameter is introduced in the batch mode module. On load/save of a batch file - this is added to the FileNameListComponent. This enables users to have multiple batch files for smaller parts of the whole processing and concatenate them fast.

image


tomas-pluskal:
Hey Robin,
As a feature, I agree it's great to have. But let's discuss the implementation.
Wouldn't it make more sense to add a "Last used..." button, for example next to the "Load.." and "Save.." buttons, which, after clicking, would show a small popup with the selection of the last used files? The popup can be very simple, similarly to the MassListParameter popup for mass list names.

In that way, there would be no need to add a new Parameter. Think about it - this is actually not a parameter of anything.

Similarly, for FileNameParameter we can add the same "Last used.." button and the list of last used files can be remembered inside the data structure of that parameter.

robinschmid:
Hey Tomas,

yeah this was a very fast job to induce a conversation about the best implementation. 
Sounds good to me. 

And how about a menu below load project
- last projects
  - p1
  - p2 
  - ...

The easiest implementation would be to add a list of files directly to MZmineConfigurationImpl right?
Then I would check if the files exist and create menu items in a sub menu of File/Last projects

What do you think?

tomas-pluskal:
Adding a menu item is possible, but I generally like the idea of keeping everything that's related to one parameter together. So I would be in favor of adding a GUI component for the last used files to each parameter that is using files.

One additional problem with this silent parameter approach: imagine you have a ParameterSet with two different filename parameters. Then it becomes unclear what this silent parameter is used for.

robinschmid:
I have added the button+popup menu as a JLastFilesButton to the FileNameComponent.
This seems to work fine.

FileNameListSilentParameter was changed to a regular Parameter (not UserParameter).
This parameter should be useful for the last batch mode files and last project files.
As we do not have a FileNameParameter in the batch mode, but two buttons (save and load) we need to have the last saved/loaded files centralised in one parameter.
Same is true for load/save project.

robinschmid:
The save/load project has no FileNameParameter and they should share the same list of last used files - so we should centralise this in the config

tomas-pluskal:
Good point with the shared list for project loading and saving.

As for the FileNameListSilentParameter, why not add the list of last filenames as an internal property of the ```BatchQueueParameter```?

robinschmid:
This was my second idea. However logically, the last used batch files do not have anything to do with the queue-parameters. They are more related to the BatchModule. 
I dont know if there is any difference in the result though. (When parameters are saved and loaded - the last used files should be as static as possible)

tomas-pluskal:
Yeah, you're right.

src/main/log/.gitignore
```diff
@@ -0,0 +1,2 @@
+/mzmine.log.1
```
tomas-pluskal: I think this should be added to the .gitignore file in the root folder

robinschmid: After changing the working directory it started to produce these files. Somehow the gitignore is not shown in the folder tree.

robinschmid:
Ok I have implemented the last projects into the MZmineConfiguration and they are displayed as menu items of a menu File/Open last project ... 

Works like a charme and it is saved to the config. 

Works for:
 - any FileNameParameter
 - Projects
 - Batch files

tomas-pluskal:
Ok, sounds good, let's merge it!

robinschmid:
Great. Let me know if you change anything or if you have some more ideas on this issue

PR #593, opened 2019-05-22
Spectral similarity functions

robinschmid:
Hey Tomas,
Hey Ansgar,

this PR adds the Weighted cosine similarity and the Composite cosine identity functions (both similar to NIST) to the spectral library search modules. Implementation is modular - so we can add more in the future. 

Another addition is the optional deisotoping of masslists before calculating the scores.

Cheers
Robin


tomas-pluskal:
Looks great, I like the modular approach to the various similarity functions.

robinschmid:
Great :)

PR #594, opened 2019-05-22
Added molecule preview and better line wrapping to DB result frame

Annexhc:
Hey Tomas,
I have further improved the spectra DB matching result frame. I have added structures if the database contains inchi or smiles. Further I have improved the line wrapping.
best wishes,
Ansgar
MSMSSpectraFrame_Update


src/main/java/net/sf/mzmine/modules/visualization/spectra/simplespectra/spectraidentification/spectraldatabase/SpectralIdentificationSpectralDatabaseTextPane.java
```diff
@@ -0,0 +1,36 @@
+package net.sf.mzmine.modules.visualization.spectra.simplespectra.spectraidentification.spectraldatabase;
```
tomas-pluskal: Please add the license header

src/main/java/net/sf/mzmine/modules/visualization/molstructure/MolStructureViewer.java
```diff
@@ -238,4 +238,8 @@ private void load3DStructure(URL url) {
     }
 
   }
+
+  public JSplitPane getSplitPane() {
```
tomas-pluskal: This function is added but it doesn't seem you actually use it anywhere?

Annexhc: oh yes, this was a leftover from another approach. I have removed it

src/main/java/net/sf/mzmine/modules/visualization/spectra/simplespectra/spectraidentification/spectraldatabase/SpectralIdentificationSpectralDatabaseTextPane.java
Annexhc: done

tomas-pluskal:
Thanks Ansgar, looks great. Additional comments below:
- please add the license header to MirrorScanWindow
- I think MirrorScanWindow should be moved to the simplespectra package, because it is not a module by itself (does not implement MZmineModule)
- As Robin mentioned in one of the conversations we had, all your modules (like OnlineDBSpectraSearchModule in net.sf.mzmine.modules.visualization.spectra.simplespectra.spectraidentification.onlinedatabase) should implement MZmineModule (not MZmineProcessingModule) and should be added to the MZmineModulesList so that their parameter set instance is created during startup and saved on shutdown)
- you should not create new instances of the ParameterSet classes like you do here (https://github.com/mzmine/mzmine2/blob/b287fd5bb3fd0dd4cba45dcd9dfabc61cbeb2d64/src/main/java/net/sf/mzmine/modules/visualization/spectra/simplespectra/spectraidentification/onlinedatabase/OnlineDBSpectraSearchModule.java#L70) but instead obtain the ParameterSet from the current MZmine configuration ```MZmineCore.getConfiguration().getModuleParameters(moduleClass)```, then run parameters.showSetupDialog(), and after that clone the parameterSet and run your module. This process ensures that the ParameterSet stored in the configuration is the latest one, and the parameter values are remembered even after MZmine shutdown

Annexhc:
Yes, I remember the conversation now! I will add all the changes.

Annexhc:
The MirrorScanWindow also needs further improvement. For now it does its job, but a lot of functionalities are still missing such as proper data export etc. I will add this in the future.

robinschmid:
I will also add some more functions to the window later.

Maybe you should put the proposed changes to another PR. (Change to a MZmineModule and the module parameters)

Annexhc:
would be okay for me as well. I can do the changes later this day or in another pull request. Tomas, you can decide.

tomas-pluskal:
I'll merge this one and you can submit the other changes in a separate PR.

PR #597, opened 2019-05-23
Spectral sim: Error fix, cancel of all tasks, match only overlapping m/z

robinschmid:
Hey Tomas,

A first small PR. 

 - fixed error in single spectra matching task
 - If main task is cancelled - all tasks are
 - Option to limit the library matching to the overlapping m/z range only
    - query spectrum (200 - 1000 m/z) and library spectrum (50 - 800 m/z) --> they will only be matched within 200-800 m/z (with the specified m/z tolerance applied to both ends) 

Still working on some things for the library submission and results visualisation.

Cheers
Robin


tomas-pluskal:
Looks good.

Issue #601, opened 2019-05-27
Open last project


tomas-pluskal:
<img width="498" alt="Screen Shot 2019-05-27 at 12 43 42" src="https://user-images.githubusercontent.com/1796692/58431713-1d28ef00-807d-11e9-89aa-39e94d015bcd.png">

How about making the Open last project menu disabled if it's empty?

robinschmid:
Sounds good

robinschmid:
Resolved

PR #602, opened 2019-05-27
Only show "open last projects" menu if n>0

robinschmid:
Resolves issue #601


src/main/java/net/sf/mzmine/desktop/impl/MainMenu.java
```diff
@@ -94,6 +94,7 @@
     add(projectMenu);
 
     lastProjectsSub = new JMenu("Open last project ...");
+    lastProjectsSub.setVisible(false);
```
tomas-pluskal: I think it's better to disable it using setEnabled(false).. disappearing menu items are confusing.

tomas-pluskal:
We can also do the same for the FileNameParameter last button.

robinschmid:
Done

Issue #598, opened 2019-05-23
Spectral library matching - Results visualizer


robinschmid:
This is a reminder that we still need to change the results visualization of the spectral library matching:

- Change metadata pane (@Annexhc )
- Change SpectralMatchIdentity (@robinschmid )
  - Save the experimental masslist + filtering steps, to visualize all filtered, matched and unmatched signals

I would suggest that we wait for this before we create next version.
My part will be done on friday, 24th 
@Annexhc  and you?

Annexhc:
Hey Robin, I hope I will finish the spectra matching by tomorrow.
One more point regarding the spectral library matching peaklist method. The identities should be sorted by matching score.

robinschmid:
resolved by #600

tomas-pluskal:
<img width="1365" alt="Screen Shot 2019-05-27 at 13 19 00" src="https://user-images.githubusercontent.com/1796692/58432796-32544c80-8082-11e9-8cec-91e2c6bd6dd6.png">

A few things I noticed:
- the title of the results window remains as "Processing..."
- what is the purpose of the task controller in the bottom of the window?
- zooming is a bit weird - if I zoom the intensity scale in the top spectrum, the bottom spectrum is unchanged, so the intensities look different
- it might be better to show relative intensity (%) instead of absolute in the spectra?
- the split between the spectra and Compound Information cannot be resized using mouse

Annexhc:
ah okay, that's weird, the processing should switch to a number of match results. The task controller was only to display the progress since it may take a while when processing a large database. In a previous version it was removed after the last task. What do you prefer? We could also remove it since the results are now added constantly from separated tasks it is not that important anymore. My first version was only using one thread ;-) But I can handle these two points for sure.
Actually I like the zooming very much as it is. I think it is good that you can zoom in each plot separately. Relative intensity makes sense!
It cannot be resized because I have switched back to a normal panel. Split pane showed kind of strange behavior. Of course we could also change back if you like it more.

robinschmid:
We could ... 
- change to relative intensities. 
- add a button/option to couple the intensitiy axes of the mirror plot

Annexhc:
@tomas-pluskal does the "Processing..." title remains only when you open the spectra results window from the peak list or does it also happen when you perform spectra matching with the spectra visualizer?

Annexhc:
never mind, Ive found the bug

Annexhc:
See PR #603

[...]

PR #603, opened 2019-05-27
Spectra matching update

Annexhc:
For #598 
Fixed spectra result window title for peak list. Removed task controller from spectra results window. Switched to rel. intensity in results frame.


src/main/java/net/sf/mzmine/modules/visualization/spectra/simplespectra/mirrorspectra/MirrorScanWindow.java
```diff
@@ -160,14 +166,14 @@ public void setScans(SpectralDBPeakIdentity db) {
       for (DataPoint dp : query[i]) {
         // not contained in other
         if (notInSubsequentMassList(dp, query, i))
-          qdata.addDP(dp.getMZ(), dp.getIntensity(), null);
+          qdata.addDP(dp.getMZ(), dp.getIntensity() / mostIntenseQuery * 100, null);
```
tomas-pluskal: This will crash if mostIntenseQuery == 0 (that could happen in principle)

robinschmid:
Thanks for working on this.

You are right with the numberformat as it is for chart visualisation. Was my bad.

tomas-pluskal:
Thanks!

PR #604, opened 2019-05-28
Kendrick update

Annexhc:
Dear Tomas,
a lot has happened in the literature regarding the analysis of polymers using Kendrick mass defect analysis. I have collected the most useful tools and implemented them to the Kendrick module. In order to ensure a clear overview I have also added a feature overview window, which appears when double clicking a data point. I have added a detailed help file with information on the literature and the mathematical formulas to explain what is going on.
best wishes,
Ansgar


tomas-pluskal:
Looks good to me, thanks!

PR #605, opened 2019-05-28
Spectral match results changes

robinschmid:
This is a first fix for the spectral match results visualisation:
- Color scale from red - to green (score = 0.5 to 1.0, respectively)
- Sizes of meta data panel and chart

A problem with the layout remains. Somehow, the meta data is visualised with larger height (also the structure). Suddenly after scrolling down, everthing changed to the actual layout. I did revalidate on every change but still...

image
image

### DIFF
--- a/src/main/java/net/sf/mzmine/chartbasics/listener/AxisRangeChangedListener.java
+++ b/src/main/java/net/sf/mzmine/chartbasics/listener/AxisRangeChangedListener.java
@@ -18,6 +18,7 @@
 
 package net.sf.mzmine.chartbasics.listener;
 
+import javax.annotation.Nullable;
 import org.jfree.chart.axis.ValueAxis;
 import org.jfree.chart.event.AxisChangeEvent;
 import org.jfree.chart.event.AxisChangeListener;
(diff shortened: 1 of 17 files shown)


robinschmid:
Ok fixed it. And the structure is a bit easier
image

Annexhc:
wow, that is some short code :-D very nice. Can you change the background of "Compound information", "Instrument information" etc. back to white? I think that would look more fancy. The gray boxes have a little gap.

robinschmid:
I actually like the gray background. It is like a header

Annexhc:
But we already have the bold text as a header. What I really don't like is the gap in between the boxes.
gap

robinschmid:
This should cover everything. 
- the MoNA json parsing is faster by keeping the reading and parsing in separate threads
- the width of the meta data panel is fixed
- there is a JCheckBoxMenuItem to toggle the zoom coupling of the two intensity axes
- title bar has auto wrapping and size now

image

robinschmid:
From my side, everything is ready to be merged.

tomas-pluskal:
Looks good, thanks!

PR #607, opened 2019-05-30
Replace Student's T-test with One-way ANOVA test

asmirn1:
Hi Tomas,

I've made a few changes to the statistical significance calculation:

- I used to determine (control and treatment) groups for statistical analysis, based on two filename patterns. In the new version, I use the sample parameters defined by going to "Project -> Set sample parameters".

- I replace Student's t-test that works only for two groups (e.g. control and treatment) with One-way ANOVA test that works for any number of groups.


tomas-pluskal:
Hi Alex,
Thanks, using the project parameters is much better, and more consistent with other modules.
I'll merge this now, but how about also adding the p-value as a column to the peak list table visualizer?

PR #610, opened 2019-06-03
Small mz format fix for GNPS library submission/creation

robinschmid:
Hey Tomas,

I thought everything was tested - but I have introduced a small bug in the library submission tool.
The m/z format should not be a String but a double.

Small fix for this.

Cheers
Robin


tomas-pluskal:
Does this break the function completely? I mean, should we release 2.41 asap with this fix?

robinschmid:
Thanks Tomas. It did not break the functionality. However, the parsing was more difficult. 
Usually it is best practice to format floating point values before they are exported to text files. As json differentiates between a number and string value - the number needs to be formatted in a different way.

Thank for releasing the quick bug-fix version

tomas-pluskal:
I mainly released the bug-fix version because the 2.40 release was broken on Java 8, due to compatibility issues in the StreamCopy class. It should be all fine now.

PR #612, opened 2019-06-05
Color palettes and mode for color blindness

robinschmid:
Hey all,

I have added a class ColorPalettes to store different colors for normal/color blind vision. The preferred vision can be changed in the MZmine preferences. 

All modules and charts can then get the colors from the ColorPalettes. 
- getPositiveColor(Vision vision) (green or blue)
- getNegativeColor(Vision vision) (red or orange)
- getSevenColorPalette(Vision vision, boolean includeBlack)
    - taken from http://mkweb.bcgsc.ca/colorblind/img/colorblindness.palettes.trivial.png

So far I have changed the colors of the charts in spectral match results and spectral database submission. From my side, I would suggest to set the default color vision to color blind (DEUTERANOPIA). These color palettes are great for nearly everyone.

```
public enum Vision {
    NORMAL_VISION, //
    DEUTERANOPIA, // green blindness, 6% male population
    PROTANOPIA, // red blindness, 1% male population
    TRITANOPIA; // blue blindness, very rare
}
```
image

### DIFF
--- a/src/main/java/net/sf/mzmine/datamodel/MassSpectrum.java
+++ b/src/main/java/net/sf/mzmine/datamodel/MassSpectrum.java
@@ -18,10 +18,9 @@
 
 package net.sf.mzmine.datamodel;
 
-import com.google.common.collect.Range;
-
 import javax.annotation.Nonnull;
 import javax.annotation.Nullable;
+import com.google.common.collect.Range;
 
 /**
  * This class represent one mass spectrum.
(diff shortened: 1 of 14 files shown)


tomas-pluskal:
Thanks Robin, that is a very thoughtful update!

robinschmid:
One of my previous mentors is color blind. People always think the only problem is red to green, but there are way more problems. Red to black, yellow, brown ....
Basically, most of the colors should have a bit of blue and different shades.

Thanks for merging :)

PR #613, opened 2019-06-07
MS-Level differentiation for DataPointProcessing

SteffenHeu:
Hi Tomas, 
this PR implements the functionality to set up different processing queues for MS^1 and MS^n scans.
DataPointProcessingModules now require the implementation of getApplicableMSLevel() to specify which ms level they are intended to be used for.

The user will get a message window if he is trying to add a MS^1 module to MSMS processing queues.
warning
(The dummy modules are not included in the PR of course.)

Furthermore, i fixed an issue that did not properly load the values of HiddenParameters into ParameterSetupDialogs (see ParameterSetupDialogs,java).

Best,
Steffen


robinschmid:
Great addition. (I haven't checked the code so far)

tomas-pluskal:
Hi @SteffenHeu 
My concern is that the dialog (as you show it in the screenshot) is becoming huge. Wouldn't it make more sense to put both MS1 and MS2 processing steps into the same tree? (e.g. the tree can have two root items, "MS1 processing queue" and "MSn processing queue").

robinschmid:
One tree would actually be better. This would also remove all the button duplicates. Maybe add a button "Add to MS^n" and enable/disable it depending on the user selection.

On double click on any item on the right, you could add this item to the last active root item on the left. Just look for the current selected item or if none is selected for the previously selected.
If the user adds one item to MS1 the double click will then add new items to MS1.

SteffenHeu:
I dont know how to implement that, because that would mean the ProcessingParameter would have an Array of DataPointProcessingQueue[] as value. Not sure how to implement the xml save/load there.

tomas-pluskal:
You don't need an array of DataPointProcessingQueue[], you just need two of them (one for MS1 and one for MSn). So you can save/load them separately, like
````
<ms1queue>
  ... processing queue data
</ms1queue>
<msnqueue>
  ... processing queue data
</msnqueue>
````

**Comment by SteffenHeu** [comment]
setup window

Modules will be added to the selected node in the left tree view. If no node is selected, they will be added to the last selected one. By default they are added to MS-any.

**Comment by tomas-pluskal** [comment]
Awesome, that looks much better!

**Comment by tomas-pluskal** [comment]
I think the "Split MS1 and MSn level processing" checkbox is unnecessary, because anyway you have a queue for "MS-any" as well as for the different MS levels. The checkbox doesn't add any new functionality in that case.

**Comment by SteffenHeu** [comment]
The checkbox is used to decide if the MS-any queue is used for every scan, or if the MS^1 and MS^n queues are used.

**Comment by tomas-pluskal** [comment]
Okay, now I understand, but I think it is quite confusing for a new user. 
How about one of these options
- When the check box is checked, make the MS-any queue disappear. And when it us unchecked, make the other two queues disappear.
- Remove the checkbox and the MS-any queue completely. If the user wants to apply processing to all scans, he/she can simply add the same processing steps to both MS1 and MSn queues. But I think most users will want to apply different processing steps to MS1 and MSn, so actually I like this option better.

**Comment by SteffenHeu** [comment]
I understand your concerns, but i think having the option to switch between a default MS-any queue and the MS^1 + MS^n queues fast is beneficial. I will try to think of something to make it clearer, which mode is enabled at the moment.

**Comment by robinschmid** [comment]
I think any of the two suggestions are good. 
Most of the time people will use different queues for MS and MS/MS.

I would rename MS^1 to MS and MS^n to MS/MS this is easier - and the help/tooltip can describe that MS/MS means MS level>1.

If you want to keep MS any, only show MS any OR MS and MS/MS depending on the state of the checkbox. (Tomas first suggestion)

**Comment by SteffenHeu** [comment]
Done:
setup window
setup window2

This is implemented in way, that adding further differentiation like ms^2 and ms^3 can be added with minimal changes.

File: src/main/java/net/sf/mzmine/modules/visualization/spectra/simplespectra/datapointprocessing/datamodel/MSLevel.java [hunk 1]
```diff
@@ -0,0 +1,37 @@
...
+
+  @Override
+  public String toString() {
+    return name;
+  }
+
+  public static MSLevel[] cropValues() {
+    MSLevel[] values = MSLevel.values();
+    MSLevel[] array = new MSLevel[values.length - 1];
+    for (int i = 0; i < array.length; i++)
```
tomas-pluskal: Use Arrays.copyOf() instead of doing this manually (the whole cropValues() function can be removed)

src/main/java/net/sf/mzmine/modules/visualization/spectra/simplespectra/datapointprocessing/datamodel/customguicomponents/HighlightTreeCellRenderer.java
```diff
@@ -0,0 +1,56 @@
...
+      setOpaque(false);
+      c.setBackground(getBackgroundNonSelectionColor());
+      return c;
+    }
+//    logger.finest("renderign value: " + value.toString() + " is mslevel node");
+    
+    boolean diffMSn = cb.isSelected();
+    
+    c.setOpaque(true);
+    c.setBackground(getNodeColor((DPPMSLevelTreeNode)value, diffMSn));
```
tomas-pluskal: How about just setting setEnabled(false) instead of using the red background?

tomas-pluskal:
Please add the license header to `ProcessingComponent.java` and `ProcessingParameter.java`. Also I think if these two classes are only used in your module, you should move them into your package (`net.sf.mzmine.modules.visualization.spectra.simplespectra.datapointprocessing`). Or do you think this parameter will be useful for other modules?

src/main/java/net/sf/mzmine/modules/visualization/spectra/simplespectra/datapointprocessing/DataPointProcessingParameters.java
```diff
@@ -20,6 +35,6 @@
       "Set the modules to be executed in the processing queue.");
 
   public DataPointProcessingParameters() {
-    super(new Parameter[] {spectraProcessing, processing});
+    super(new Parameter[] {processing, spectraProcessing});
```
tomas-pluskal: Maybe use more self-explanatory names than `processing, spectraProcessing`? 
Like `processingQueue, processingEnabled`.

src/main/java/net/sf/mzmine/modules/visualization/spectra/simplespectra/datapointprocessing/datamodel/customguicomponents/HighlightTreeCellRenderer.java
robinschmid: Can you do setEnable(false) on a tree item?
This would be the best way.

src/main/java/net/sf/mzmine/modules/visualization/spectra/simplespectra/datapointprocessing/datamodel/MSLevel.java
SteffenHeu:
```
for (MSLevel mslevel : MSLevel.cropValues())
      value.setQueue(mslevel, getProcessingQueueFromNode(getNodeByMSLevel(mslevel)));
```
Isn't it cleaner to use a method of the enum than Arrays.copyOf() in every for loop like this?

src/main/java/net/sf/mzmine/modules/visualization/spectra/simplespectra/datapointprocessing/datamodel/customguicomponents/HighlightTreeCellRenderer.java
SteffenHeu: Tree nodes dont have that functionality, but can make a one, of course. You still need a new renderer though.

tomas-pluskal: Every JComponent has setEnabled(), so you can just call c.setEnabled(false) here.

src/main/java/net/sf/mzmine/modules/visualization/spectra/simplespectra/datapointprocessing/datamodel/MSLevel.java
tomas-pluskal: Okay.

src/main/java/net/sf/mzmine/modules/visualization/spectra/simplespectra/datapointprocessing/datamodel/customguicomponents/HighlightTreeCellRenderer.java
SteffenHeu: https://github.com/mzmine/mzmine2/blob/6be7cc01dcbbe5f53ea7bc167ed3f22a9442bece/src/main/java/net/sf/mzmine/modules/visualization/spectra/simplespectra/datapointprocessing/datamodel/customguicomponents/DisableableTreeCellRenderer.java#L50

Thats what I'm doing, but i did not find the fuctionality to store that information in DefaultMutableTreeNode, thats why i made the DisableableTreeNode (it's toungebreaker, i know :D)

tomas-pluskal:
Looks good to me! Thanks!

PR #614, opened 2019-06-11
Bugfixes DataPointProcessing

SteffenHeu:
Hi Tomas, i fixed a few small bugs I noticed.

1. After setting new parameters, the processing was executed, but the previous results were not removed.
2. A null pointer exception occurred, when the last selected node and currently selected node were both null and a data point did not have a result at all and a remove method was called. 
3. Furthermore, i added Nonnull and Nullable annotations to ProcessedDataPoint.

Best regards
Steffen


tomas-pluskal:
Looks good, thanks!

Issue #615, opened 2019-06-11
ADAP chromatogram deconvolution hangs


tomas-pluskal:
It seems setting the "RT wavelet range" parameter to a range with the same min and max (like 0.1 - 0.1) causes the method to hang with high CPU use. 

@du-lab @asmirn1 Could you take a look at that?

asmirn1:
Sure, I'll fix it.

asmirn1:
Hi Tomas,

I've pulled the latest version of MZmine 2, and now it doesn't compile for me. I'm using Java 8. Here is the output:
```shell
mvn clean install
```

[INFO] Scanning for projects...
[INFO] 
[INFO] ----------------------< io.github.mzmine:mzmine2 >----------------------
[INFO] Building MZmine 2.40.1
[INFO] --------------------------------[ jar ]---------------------------------
[WARNING] The POM for org.rsc.chemspider:chemspider-api:jar:1.0.0 is missing, no dependency information available
[WARNING] The POM for mzmine:macspecificsetup:jar:2.7 is missing, no dependency information available
[WARNING] The POM for mzmine:jmprojection:jar:1.0 is missing, no dependency information available
[WARNING] The POM for mzmine:gslibml:jar:0.0.1 is missing, no dependency information available
[WARNING] The POM for mzmine:visad:jar:20131024 is missing, no dependency information available
[WARNING] The POM for java3d:j3dcore:jar:1.6.2 is missing, no dependency information available
[WARNING] The POM for java3d:j3dutils:jar:1.6.2 is missing, no dependency information available
[WARNING] The POM for java3d:vecmath:jar:1.6.2 is missing, no dependency information available
[WARNING] The POM for de.unijena.bioinf.ms:chemistry_base:jar:3.1.4-SNAPSHOT is missing, no dependency information available
[WARNING] The POM for de.unijena.bioinf:java-utilities:jar:1.0-SNAPSHOT is missing, no dependency information available
[WARNING] The POM for gurobi:gurobi-linux64:jar:7.5.2 is missing, no dependency information available
[WARNING] The POM for cplex:cplex:jar:12.8 is missing, no dependency information available
[WARNING] The POM for de.unijena.bioinf.ms:chemical_db_oss:jar:1.1.3 is invalid, transitive dependencies (if any) will not be available, enable debug logging for more details
[INFO] 
[INFO] --- maven-clean-plugin:2.5:clean (default-clean) @ mzmine2 ---
[...]

PR #618, opened 2019-06-13
Bugfix for #615

asmirn1:
Fix a bug in ADAP when the upper and lower wavelet scales are the same.
Update the ADAP version.


tomas-pluskal:
Thanks, Alex. 

Don't you think it might be a good idea to update the parameter class in MZmine to prevent the user from setting the same upper and lower scales?

asmirn1:
Sure. I guess the best way to do it would be using `UserParameter.checkValue()`? Let me figure out how you are using this method.

asmirn1:
I've created a new class DoubleNonEmptyRangeParameter that extends DoubleRangeParameter. Will this work?

tomas-pluskal:
It seems like an overkill to create a new class that duplicates the existing functionality. Why not add a simple property to the existing DoubleRangeParameter?

asmirn1:
I've updated DoubleRangeParameter and removed DoubleNonEmptyRangeParameter

tomas-pluskal:
Looks good, thanks!

PR #622, opened 2019-06-24
Update ADAP Aligner name and its help file

asmirn1:
I've renamed ADAP Aligner -> ADAP Aligner (GC) and made a small change in the help file.


tomas-pluskal:
Thanks, Alex!

I think the information in the help is not very clear to people who are newcomers to MZmine. How about adding a sequence of steps (modules) that the user should run in order to get the data ready for the ADAP GC aligner? (either using CAMERA or using the deconvolution modules)

asmirn1:
Hi Tomas,

I've added an example of the workflow to use with ADAP Aligner (GC). Please, take a look.

Alex.

tomas-pluskal:
Looks good, thanks Alex!

Issue #620, opened 2019-06-23
Feature request: rename in Peak list drop down menu


photocyte:
In the current Peak list interface (MZmine2 v2.40.1), it is possible to rename Peak lists.

<img width="393" alt="Screen Shot 2019-06-23 at 9 51 38 AM" src="https://user-images.githubusercontent.com/12517705/59977228-cee40e80-959c-11e9-8ce9-6af5871ce6a9.png">
(editing box shown with red arrow)

But, the only way I am aware to get the Peak list name text box to be activated for editing, is to single click somewhere at the right end of the Peak list name.

It can be difficult to click in the exact right spot to get this text box editing to activate.  I'd suggest that an option be added to the right click drop down menu which activates this text box editing for the Peak list as well.  E.g. a "Rename Peak list" option.

<img width="481" alt="Screen Shot 2019-06-23 at 9 42 15 AM" src="https://user-images.githubusercontent.com/12517705/59977257-0a7ed880-959d-11e9-9c94-efa28ecd5e77.png">

Annexhc:
Have you tried single click on the feature list + pressing F2?
That works for me. If it does not work on your system I can add the rename function to the pop up menu.

photocyte:
It works for me, in a strange way (on MacOS).  I can simultaneously left click + F2, and after a solid 600ms delay, the text box will become editable.  Similarly, if I left click, then press F2, the text box will become editable, after a similar ~600ms delay.  Pressing F2 alone doesn't do anything.  Is this behavior documented anywhere?

Annexhc:
Sorry, I have no experience with MacOS and I cannot test it. On linux I have normal behavior, no delays or something similar. I think @tomas-pluskal  uses MacOS. Maybe he can test it.

tomas-pluskal:
For me the editing always works, no matter where I click (it doesn't need to be near the right end). But adding the Rename item to the popup menu certainly won't hurt anything.

Annexhc:
Okay, I can do it.

PR #626, opened 2019-06-26
Added feature list renaming to popup menu

Annexhc:
Hi Tomas,
I have added feature list renaming to the popup menu as requested in #620.
best Ansgar


tomas-pluskal:
Thanks Ansgar!

I think it is actually not necessary to introduce a new dialog. Isn't it simpler to just activate the native editing function of the JTree? (the same as what happens after single click on the peak list name)

Also, for consistency we should add the same Rename menu item to the raw data file list, too.

Annexhc:
Yes you are right, it is much simpler, but I thought it might be a bit more fancy to have a dialog. Anyway, I don't really care, since I will use the F2 key in the future :-P

tomas-pluskal:
Thanks a lot, Ansgar! :)

Issue #627, opened 2019-06-27
Problem with cosine value within ID-tag during join alignment


oligegilo:
I am reporting a problem with MZmine 2 version 2.40.1
My operating system is 64-bit.

I'm comparing peaklists of different samples that have been processed with a local spectral database based on MS2 data. During Join Aligner step I chose the option "Require same ID" so that identified features will be aligned based on their prior MS2 similarity and not just m/z and RT. So far so good, but there is a problem to merge similar features that have been identified to be the same compound but with different cosine scores - those will be treated as two different identifiers and not merged into one feature during the alignment step (see picture below). 
I assume if you remove the cosine value from the ID-tag then the merge based on just the name would work. Anything I can do right now to avoid that? Thanks!

![Alignment](https://user-images.githubusercontent.com/43857712/60277972-e3a70600-98fe-11e9-923b-1d1295866bb0.JPG)

oligegilo:
Ok quick update from my side: I edited the ID labels myself within the peaklists before the merging step and then it works as expected - still editing the labels per hand is quite laborious when dealing with big data..

![Alignment_2](https://user-images.githubusercontent.com/43857712/60278820-5ebcec00-9900-11e9-9b2a-f0daf3f42a9e.JPG)

Annexhc:
We could move the score to the comment.

oligegilo:
That would be perfect! I'm stuck right now with my analysis because of that. I was able to export the peak list as mzTab and delete the cosinus value from the ID label via a shell command but when I import the mzTab files again the actual MS2 data is not linked to the peak list anymore, which makes a .mgf export impossible...anyways, your suggestion would be the easiest solution here.

Annexhc:
okay, if it is really urgent I can send you a custom version for your requirements. We can then still decide if we put these changes to the main branch of MZmine. You can contact me: ansgar.korf@wwu.de

robinschmid:
Be aware that if you acquire data for n samples, alignment will only be correct on all samples if you aquire an MS2 scan in each sample for this feature (specific analyte ion).

We still need to find a way to compare library entries (you could have multiple MS2 library spectra in the library from different instruments and fragmentation energies...)

robinschmid:
Maybe it would be better to introduce a MS2 cosine similarity to the join aligner and to only align features with similar MS2 spectra.
@tomas-pluskal @Annexhc what do you think?

PR #631, opened 2019-06-30
Moved cos scorce to comment

Annexhc:
Hi Tomas,
this pull request fixes issue #627.
Ansgar

### DIFF
--- a/src/main/java/net/sf/mzmine/modules/peaklistmethods/identification/spectraldbsearch/PeakListSpectralMatchTask.java
+++ b/src/main/java/net/sf/mzmine/modules/peaklistmethods/identification/spectraldbsearch/PeakListSpectralMatchTask.java
@@ -19,6 +19,7 @@
 package net.sf.mzmine.modules.peaklistmethods.identification.spectraldbsearch;
 
 import java.io.File;
+import java.text.DecimalFormat;
 import java.text.MessageFormat;
 import java.util.List;
 import java.util.logging.Level;
@@ -328,6 +329,9 @@ public class PeakListSpectralMatchTask extends AbstractTask {
     // add new identity to the row
     row.addPeakIdentity(new SpectralDBPeakIdentity(getScan(row), massListName, ident, sim, METHOD),
         false);
+    // add score to comment
+    DecimalFormat COS_FORM = new DecimalFormat("0.000");
+    row.setComment("cos=" + COS_FORM.format(sim.getScore()));
   }
 
   public int getCount() {
--- a/src/main/java/net/sf/mzmine/util/spectraldb/entry/SpectralDBPeakIdentity.java
+++ b/src/main/java/net/sf/mzmine/util/spectraldb/entry/SpectralDBPeakIdentity.java
@@ -18,7 +18,6 @@
 
 package net.sf.mzmine.util.spectraldb.entry;
 
-import java.text.DecimalFormat;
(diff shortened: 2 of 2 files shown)


tomas-pluskal:
Thanks, Ansgar.
Now looking at the code, I have some doubt whether this is the best solution. Each peak list row can have multiple identities assigned to it, but only one comment field. If during the spectral matching two possible hits are assigned to the peak list row, the comment field will create confusion, don't you think?

Annexhc:
Oh yeah that is really not great. What is your suggestion? We could add all identities to the comment as well. This could result in very long comments thought.

tomas-pluskal:
I think a better solution is to make sure that two PeakIdentity instances are considered same when they refer to the same library spectrum, but disregarding the cosine similarity. That means overriding the equals() method in SpectralDBPeakIdentity, and making sure that the Join Aligner actually uses the equals() method to check whether two identities are the same.

Annexhc:
Hey Tomas, sorry for the delay. Your solution sounds way better. I will think about the implementation.

Still I have one concern. What if the database contains more than one spectrum for the same compound, e.g. in GC-MS one deconvoluted, the other not? Or more likely, two spectra for the same compound acquired with different instruments (QTOF vs. Orbitrap)?

What do you think about adding a new parameter to the PeakIdentity, something like "Identiification method". There we could store the applied identification module and other related information like the cos score. 
image

robinschmid:
We need a multi layered comparison of the library hits. 
The best description should be the structure as InChI or SMILES. @tomas-pluskal do you know of a structure comparison method (e.g., in the CDK)?

1. Compare structure (with or without stereochemistry)
2. Compare mol. formula???
3. Compare name (ignoreCase; which can be very different and unprecise depending on the library entry)

tomas-pluskal:
The spectral matching module is supposed to match spectra, not compounds. So I think if the database contains multiple spectra of the same compound, these should be still reported as separate hits.

Perhaps the best solution would be to extend the data model, so for each PeakListRow, we can record
- chemical structure, if known (this can be shown in the peak list table)
- chemical formula (of the neutral molecule)
- ionization type/adduct
- database hits/prediction results with scores - here we can aggregate the results of the different search methods. The user can then pick their favorite result (compound) and transfer it to the "chemical structure" field

@Annexhc We have the property "Identification method" in Peak Identity (https://github.com/mzmine/mzmine2/blob/master/src/main/java/net/sf/mzmine/datamodel/PeakIdentity.java#L36), but the dialog could certainly be improved

Annexhc:
Okay Tomas, I understand your point. The extension of the data model would be great. I think @robinschmid has done a lot of that already in his ion identity branch.

robinschmid:
They should definitely be reported separately. A way of comparing spectral library matches or structures in general would still be very helpful (e.g., to find the best common structure, to compare IDs in join aligner (see issue #627), ...)

I already changed the structure in one of my development branches for ion identity networking:
- group ID (grouped features)
  - ion identity network ID (ions of the same neutral molecule)
  - Ion identity (adduct)
    - neutral mass
    - molecular formula
  - best average formula for all ions (for an ion identity network)

Annexhc:
I think, I will close this PR for now. I will work again on issue #627 in the future when Robin has implemented his changes to the data model. As pointed out in this discussion, the changes I did so far have to many drawbacks.

Issue #583, opened 2019-05-13
MS/MS library export


tomas-pluskal:
<img width="1230" alt="Screen Shot 2019-05-13 at 14 49 47" src="https://user-images.githubusercontent.com/1796692/57646405-7c233a00-758e-11e9-9014-ac3424c714c7.png">

@robinschmid 
I think this export module needs some improvement in terms of documentation and tooltips.
Try to look at this window from the perspective of a first-time user. There is no "Help" button, and fields like "Local file" have no explanation about what kind of file is expected there. "username" and "password" for what? Etc.

Also, can we use the MassListParameter here for the mass list name?

tomas-pluskal:
If I uncheck the "Submit to GNPS" checkbox, it still requires the username & password.

robinschmid:
Thanks for reporting the wrong behaviour. Now I remember that Ansgar already pointed this out. 
This module will definitely see some development in the future. 
For now I will fix the username and password and try to give some tooltips.

How can I add a help menu to this dialog?

tomas-pluskal:
It's probably best to add a Help button just like other ParameterSetupDialogs have, to keep it consistent.

tomas-pluskal:
Generally speaking, for future compatibility and consistency among modules, it would be much better if this window could be implemented as a standard ParameterSetupDialog. For example, the GNPS export can be organized as an OptionalModuleParameter. The scan preview section can be implemented in the same way as the mass detection preview.

tomas-pluskal:
@robinschmid I would like to release v2.40 with all the recent additions, but I think this issue deserves some attention before we release it to the public. Will you have some time to work on it?

robinschmid:
Definitely, this was the next thing on my list. I am trying to finish this by tomorrow or wednesday. 
(After the last used files PR #592 )

robinschmid:
resolved by #595

tomas-pluskal:
@robinschmid The spectral library export still seems quite buggy:

<img width="1210" alt="Screen Shot 2019-06-11 at 22 24 15" src="https://user-images.githubusercontent.com/1796692/59319185-b9670e80-8c97-11e9-81e2-1263550804e3.png">

When I click Check:

<img width="248" alt="Screen Shot 2019-06-11 at 22 24 57" src="https://user-images.githubusercontent.com/1796692/59319207-cc79de80-8c97-11e9-95fe-c889fbb935b7.png">

When I click Submit:

[...]

PR #632, opened 2019-07-02
Fixed sorting and adduct textbox in library submission

robinschmid:
Hey Tomas,

this is a small fix for the spectral library export/submission. (issue #583)
The MS level initialisation was wrong and led to the weird behaviour of the adduct textbox. 

Cheers
Robin


tomas-pluskal:
Thanks, Robin!

Issue #630, opened 2019-06-29
 Feature request: Spectral deconvolution


Annexhc:
Hi guys,
this request goes to @asmirn1. Can you add the choice "most intense" to the parameter "Choice of Model Peak based on" for your spectral deconvolution modules? For me it is quite hard to predict the "sharpness" of a peak over a big data set, because the peak shape heavily depends on the column condition. The latter of course can change a lot after 100 GC-MS runs. Anyway, as the ionization should be very constant between runs, using the most intense peak of the cluster as the model peak is maybe more easy to predict for the user and/or more reliable for alignment. I would have changed it myself for testing, but I think a lot of the code is in one of your Du-Lab packages. What do you think?
best wishes,
Ansgar

asmirn1:
Yes, I think we can do it. In my experience, sharpness is quite similar to the maximum intensity. Unless there is a strong baseline level in your signal - in the latter case, sharpness might be a preferred choice. Let me talk to @duxiuxia about adding "most intense" to the parameter "choice of model peak based on".

Annexhc:
Hey Alex, that would be great! Since it is just one more option it would not hurt anyone thought.

asmirn1:
Ansgar,

I've created a pull request. Let me know if it works for you.

Alex.

Annexhc:
Hey Alex,
thank you very much for your fast response! I will test it right away and let you know about it.
best,
Ansgar

Annexhc:
Hello Alex,
thanks for the implementation again. After some testing I think you were right. The results using intensity are really similar to sharpness. Anyway, since it is an addition, which is quite easy to understand, we should keep the option.
best,
Ansgar

PR #633, opened 2019-07-02
Add an Intensity option for parameter Choice of Model Peak based on in Spectral Deconvolution

asmirn1:
This is the feature requested by Ansgar #630


tomas-pluskal:
Sounds good.

PR #644, opened 2019-07-10
Remove extra zero intensities from spectra visualizer

robinschmid:
This is a small optimisation of the spectra visualizer. Extra zeros are removed from the data sets (this speeds up visualisation).

For one of my qTOF scans from mz 100-2000 I had 400k data points. Now they are down to 2000.

If the intensities are 
01352100000000000000000000485200000000
the result is 
01352100485200

of course with the corresponding m/z values. Works like a charme.


tomas-pluskal:
The disadvantage of this approach is that when the user clicks the "Toggle displaying data points in continuous mode" button in the toolbar, it will not show the zero data points. So it is a bit misleading, because it does not show the raw data in the original "raw" state, but it shows the data somewhat pre-processed.

Anyway, I guess if you have ~400k zero data points in a single scan, that is a huge performance loss, so it is worth doing this even as a compromise.

robinschmid:
Thats true... However, I do not know how many people count zero intensity data points in a spectrum

PR #645, opened 2019-07-10
Change ScanSelection for single scan filter like smoothing

robinschmid:
I was trying to apply smoothing (single scan filter) to MS/MS scans - the result was that the filter just ignored the MS/MS scan and left me with an empty raw data file. 

Therefore, this PR adds the scan selection to the single scan filter. Not everyone acquires MS/MS scans in centroid mode - so scan smoothing should be applicable to all types of scans.

cheers
Robin

### DIFF
--- a/src/main/java/net/sf/mzmine/parameters/parametertypes/selectors/ScanSelection.java
+++ b/src/main/java/net/sf/mzmine/parameters/parametertypes/selectors/ScanSelection.java
@@ -126,7 +126,16 @@ public class ScanSelection {
 
   public boolean matches(Scan scan) {
     // scan offset was changed
-    int offset = scanNumberRange == null ? 1 : scanNumberRange.lowerEndpoint();
+    int offset;
+    if (scanNumberRange != null)
+      offset = scanNumberRange.lowerEndpoint();
+    else {
+      // first scan number
+      if (scan.getDataFile() != null && scan.getDataFile().getScanNumbers().length > 0)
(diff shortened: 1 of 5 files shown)


robinschmid:
The ScanSelection was changed. Now we have a new matches(Scan) method. Due to this change the offset calculation was also changed a bit. However, the implementation of offset seemed to be a bit random.

What do you think?

src/main/java/net/sf/mzmine/parameters/parametertypes/selectors/ScanSelection.java
```diff
@@ -101,112 +96,78 @@ public String getScanDefinition() {
...
+  }
 
-      if ((polarity != null) && (!polarity.equals(scan.getPolarity())))
-        continue;
 
-      if ((spectrumType != null) && (!spectrumType.equals(scan.getSpectrumType())))
-        continue;
+  public boolean matches(Scan scan) {
+    // scan offset was changed
+    int offset = scanNumberRange == null ? 1 : scanNumberRange.lowerEndpoint();
```
tomas-pluskal: Actually in case scanNumberRange==null, I think the offset should not be set to 1, but to the first scan number in the file. That is not necessarily 1.

robinschmid: In that case, I have changed it to 
offset = scan.getDataFile().getScanNumbers()[0];

robinschmid:
Ready from my side

PR #646, opened 2019-07-10
Export of spectra

robinschmid:
Hey Tomas,

this PR contains some changes to the spectra export (to mgf, msp, txt, mzml) which is implemented into the SpectraVisualizer.
Instead of creating a new JFileChooser every time, I have added a parameter set and Module to store the settings. Furthermore, the user can specify a mass list to export mass lists rather than full profile spectra. 
A new popup menu item allows to select multiple scans to export all of them into one file. This can also be used in combination with SIRIUS.

The MSLEVEL was added to the mgf export.

image

Cheers
Robin


tomas-pluskal:
Looks great, Robin!

I think it would make sense to move the export functions into their own package (i.e., not under visualization.spectra.simplespectra).

robinschmid:
- it was moved to the rawdatamethods
- the module was also added to the main menu (plus ScanSelection and raw file selection)

Ready from my side

tomas-pluskal:
Awesome, thanks!

Issue #319, opened 2017-06-21
ADAP deconvolution - no peaks?


tomas-pluskal:
Hi,
I loaded the sample MZmine project (https://github.com/mzmine/sample_data/releases/download/v2.3/sample-2.3.mzmine) and tried to run the ADAP chromatogram deconvolution module, but it does not seem to be detecting any peaks. See the attached details.
<img width="1356" alt="screen shot 2017-06-20 at 23 14 49" src="https://user-images.githubusercontent.com/1796692/27365529-8dc21ad4-560e-11e7-81c5-7170cf1d7181.png">

lfnothias:
Hello Tomas,
I was playing around with ADAP for LC and it works really well !
Try the following:
- S/N 5
- peak duration of 0.005 and 0.5
- using a RT wavelet range of 0.01 and 0.2. It should work better. [This parameter is really sensitive !]

RicardoMBorges:
Can I get this opportunity to ask exactly what is the parameter wavelets from the ADAP algorithm? ...I read a paper on wavelets form chromatogram, but no success in understanding it.

asmirn1:
Hi Ricardo,

I'm not sure that I understand your question. Which parameter are you referring to? 

If you want to read more about ADAP Chromatogram Deconvolution algorithm, then you can take a look at this paper [https://pubs.acs.org/doi/10.1021/acs.analchem.7b00947](https://pubs.acs.org/doi/10.1021/acs.analchem.7b00947)

Also, me and @duxiuxia might be able to answer your questions.

xiaodfeng:
I come across similar problem.

asmirn1:
@xiaodfeng 

Can you explain the problem and share a sample data file, where the ADAP deconvolution doesn't work?

RicardoMBorges:
I didn’t find a situation where it didn’t work but I cannot define a good value for the wavelets because I don’t understand its definition. So, I keep trying random values until small peaks are found and baseline is rejected as peaks.
Thanks

xiaodfeng:
@asmirn1 I got no error when use MZmine 2.40.1 on my laptop. However, there are only limited peaks identified by ADAP deconvolution module, which is suspicious. 
1. The Profile dataset is obtained by UPLC-Orbitrap QE.
2. Use msconvert's peak picking function to obtain Centroid dataset. This Centroid dataset has been shared to your email: asmirno1@uncc.edu.
[...]

PR #650, opened 2019-07-15
Add an error message when both MS1 and MS2 signals are used to build …

asmirn1:
As discussed in #319,

I add an error message when both MS1 and MS2 signals are used to build chromatograms.


tomas-pluskal:
I think stopping the task with an error is too harsh. The user might have a reason to do that (e.g., if the MS levels are not annotated correctly in the data), so it is better to just display a warning when the module is executed. 
Please follow the same method as here: https://github.com/mzmine/mzmine2/blob/master/src/main/java/net/sf/mzmine/modules/rawdatamethods/peakpicking/massdetection/MassDetectionParameters.java#L129

asmirn1:
I've replaced the error message with a warning

tomas-pluskal:
Thanks @asmirn1 !

PR #651, opened 2019-07-15
fix from merge_and_cor branch

kaibioinfo:
Fix a NullpointerException which might occur if there is no preceding or succeeding MS1 scan for an MS/MS.


tomas-pluskal:
Thanks, Kai!

PR #653, opened 2019-07-16
Van krevelen update

Annexhc:
Hey Tomas,
I have fixed a bug in the Van Krevelen diagram module. Some labels were incorrectly assigned. I have added a filter, that only feature list rows with molecular formulas are plotted. I have slightly modified the molecular formula prediction module, because the predicted formula was only assigned as identity name and not as formula. 
I have also added the FeatureOverviewWindow to he Van Krevelen Plots.
best wishes,
Ansgar


tomas-pluskal:
Looks good, thanks!

Issue #640, opened 2019-07-08
Suggestion: Pre-define model peak for ADAP spectral deconvolution!


minzaroma:
Hey everyone,

I am very fond of MZmine and the ADAP modules for processing my GC-MS data. Unfortunately there seem to be some significant troubles with quantification! When using ADAP spec deconvolution, identical peaks in different samples will often have different model peaks assigned. The outcome is a little better when choosing "sharpness", but in the end it's still very inconsistent overall. As a result, each peak area will be calculated on a different basis.

The issue can be seen here: (1) the mid (blue) row has a supposed "outlier" with 10x larger area; (2) changing the m/z value to 248 resolves the problem (the samples' concentration is about the same).
![mzmgit2](https://user-images.githubusercontent.com/52671757/60833977-d243e080-a1bf-11e9-9d36-5af7380683a9.png)

I sieved through the aligned peak table in order to match each result manually, so that each peak is represented by one and the same fragment. After finishing all the smooth workflow, well … that's some kind of a BUMP. ;)

I thought about 2 possible solutions:
1) If identification is used, the user could pre-define specific model peaks for each component in a data file. The main advantage would be that isomers or "split" components that appear in several rows may be quantified more easily. Also, alignment is not necessary (which is not feasible in all cases).
2) If alignment is used, the ADAP aligner could assign the same model peak (e.g. the sharpest or highest one) for each result row. This would be a convenient solution even without identification of the samples.

Unfortunately I don't have programming skills myself, but would be very glad if you maybe could implement such a function in the future!!!

Thanks & best wishes,
Caro

asmirn1:
Hi Caro,

Actually, I've already been working on the second solution in your list, but didn't test it yet. Would you mind testing it for me if I give you a version of MZmine with the new alignment algorithm?

Alex.

minzaroma:
That would be perfect, thanks a lot!!

Annexhc:
Alex, you could also merge the changes with the most intense model peak in that branch. Maybe Caro finds it useful. 
Great news that you are working on an improved alignment step. This seems to be the bottle neck of the ADAP GC-MS workflow at the moment. I would love to use it as well :-)

asmirn1:
@Annexhc @minzaroma 
In the new ADAP Aligner, I find the most common model peak among aligned features and scale other model peaks accordingly. I'll let you know when it's ready.

asmirn1:
[...]

PR #655, opened 2019-07-18
Setting the same m/z values for aligned features in ADAP Aligner. #640

asmirn1:
This change in ADAP Aligner selects the most common m/z of the aligned compounds, so all aligned compounds have the same m/z value and the correct height and area values.

See the discussion #640.


tomas-pluskal:
Looks good, thanks!

Issue #664, opened 2019-07-30
ADAP Chromatogram Builder reports progress inaccurately


tomas-pluskal:
The ADAP Chromatogram Builder reports the progress ("% done") as 100%, but afterwards it still takes a looooooooong time to finish creating the peak list. The getFinishedPercentage() method of the task should be updated to report progress more accurately.

asmirn1:
Okay, I'll work on it.

PR #665, opened 2019-07-30
Fixing the progress bar #664

asmirn1:
Does this progress work better?

### DIFF
--- a/src/main/java/net/sf/mzmine/modules/masslistmethods/ADAPchromatogrambuilder/ADAPChromatogramBuilderTask.java
+++ b/src/main/java/net/sf/mzmine/modules/masslistmethods/ADAPchromatogrambuilder/ADAPChromatogramBuilderTask.java
@@ -23,7 +23,7 @@ package net.sf.mzmine.modules.masslistmethods.ADAPchromatogrambuilder;
 import com.google.common.collect.TreeRangeSet;
 import com.google.common.collect.RangeSet;
 
-import java.util.Arrays;
+import java.util.*;
 import java.util.logging.Logger;
 
 import net.sf.mzmine.datamodel.DataPoint;
@@ -47,10 +47,7 @@ import net.sf.mzmine.util.SortingProperty;
 
 import net.sf.mzmine.util.DataPointSorter;
 import com.google.common.collect.Range;
-import java.util.List;
-import java.util.ArrayList;
-import java.util.HashMap;
-import java.util.Iterator;
+
 import java.lang.*;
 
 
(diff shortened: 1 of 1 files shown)


src/main/java/net/sf/mzmine/modules/masslistmethods/ADAPchromatogrambuilder/ADAPChromatogramBuilderTask.java
```diff
@@ -266,19 +261,18 @@ public void run() {
     // Stopwatch stopwatch3 = Stopwatch.createUnstarted();
 
 
-    processedPoints = 0;
-    totalPoints = simpleAllMzVals.length;
+    progress = 0.0;
+    double progressStep = 0.5 / simpleAllMzVals.length;
```
tomas-pluskal: any chance simpleAllMzVals.length might be 0 at this point?

src/main/java/net/sf/mzmine/modules/masslistmethods/ADAPchromatogrambuilder/ADAPChromatogramBuilderTask.java
```diff
@@ -370,15 +364,19 @@ else if (toBeLowerBound.equals(toBeUpperBound) && plusRange != null) {
     // System.out.println("making new chrom (ms): " + stopwatch2.elapsed(TimeUnit.MILLISECONDS));
 
     // finish chromatograms
-    Iterator<Range<Double>> RangeIterator = rangeSet.asRanges().iterator();
+    Set<Range<Double>> ranges = rangeSet.asRanges();
+    Iterator<Range<Double>> RangeIterator = ranges.iterator();
 
     List<ADAPChromatogram> buildingChromatograms = new ArrayList<ADAPChromatogram>();
 
+    progressStep = 0.5 / ranges.size();
```
tomas-pluskal: any chance ranges.size() might be 0 at this point?

src/main/java/net/sf/mzmine/modules/masslistmethods/ADAPchromatogrambuilder/ADAPChromatogramBuilderTask.java
asmirn1: I guess it's possible. I wouldn't expect it to cause any errors since `progressStep` would never be assigned to `progress` in this case. But if you want I can set `progressStep` explicitly to 0.0 for example.

tomas-pluskal: This line will throw an ArithmeticException in case simpleAllMzVals.length == 0.
Same with ranges.size() below.
You should always explicitly check for zero when doing a division by an integer.

asmirn1:
I've added checks for zero-size arrays

tomas-pluskal:
Thanks, Alex!

PR #666, opened 2019-07-31
Fix of ADAP Aligner progress bar

asmirn1:
This version of ADAP contains a fix of ADAP Aligner progress bar. In the previous version, the progress would go beyond 100%.


asmirn1:
Tomas,

Do I need to update `build.gradle` too?

tomas-pluskal:
You can update the build.gradle as well, but you don't need to. I will update all the references in the gradle script once we make the final switch from maven to gradle.

PR #667, opened 2019-07-31
Options to add retention time and ANOVA p-value to MSP files

asmirn1:
Hi Tomas,

This is a small change to MSP Export module. It adds two parameters that users can select to add Retention Time and ANOVA p-value to MSP files.

Previously, each MSP file would contain retention time. I would like to add an option to output ANOVA p-value as well.


tomas-pluskal:
Sounds good to me!

PR #671, opened 2019-08-06
Move alignment score from Comments to PeakInformation

asmirn1:
Previously, ADAP Aligner module would use `PeakListRow.setComment()` to save the alignment score. This makes the aligned peaklist to look somewhat cluttered. See the figure.

alignment_score

With this update, I save the alignment score into `PeakInformation` instead.

Tomas, I could create a separate variable of `PeakListRow` to store the alignment score if you prefer the latter.


tomas-pluskal:
Sounds good, for the time being. The peak list row data structure will need some redesign, probably for MZmine 3.

asmirn1:
Speaking of redesign, it would be nice to add a significance test p-value to `PeakListRow` as well.

PR #674, opened 2019-08-11
Added swing to PDF, EMF, EPS for SpectralMatchResultsWindow

robinschmid:
Hey Tomas,

this is a small addition to the spectral match results.
I have added the buttons ALL, PDF, EMF, EPS to export the results panel (mirror chart and meta data)  to one or all of these formats. 

Cheers
Robin


tomas-pluskal:
Thanks, Robin, looks good to me!

Issue #687, opened 2019-08-29
Isotope pattern preview window crashes


tomas-pluskal:
The Isotope pattern preview window seems to have difficulty with simple chemical formulas like "C1": 

```
[21:17:29|WARNING|net.sf.mzmine.util.FormulaUtils]: Formula invalid! Element(s) C1 do not exist.
[21:17:29|FINE|net.sf.mzmine.modules.tools.isotopepatternpreview.IsotopePatternPreviewDialog]: Invalid input or Element == "" or invalid elements.
[21:17:29|FINE|net.sf.mzmine.modules.tools.isotopepatternpreview.IsotopePatternPreviewDialog]: updateParameters() failed due to invalid input.
[21:17:29|WARNING|net.sf.mzmine.modules.tools.isotopepatternpreview.IsotopePatternPreviewDialog]: updateWindow() failed. Could not update parameters or parameters are invalid.
Please check the parameters.
```

tomas-pluskal:
@SteffenHeu

SteffenHeu:
I will look into it as soon as I can. I'm sure it's a simple check that fails, so it should be a quick fix. I am having trouble setting up java 12 and and gradle though and I am leaving for a research internship tomorrow, so I'm quite busy at the moment.
I am confident, that i can do a fix within the next couple of days, though.

tomas-pluskal:
Fixed in version 2.51

PR #692, opened 2019-09-01
Isotope pattern preview bugfix (#687)

SteffenHeu:
Hi Tomas,
this should fix the Isotope pattern preview bug (#687 ). 
Let me know if you encounter any further issues.

Best regards
Steffen

### DIFF
--- a/conf/.gitignore
+++ b/conf/.gitignore
@@ -1,0 +1,1 @@
+/config.xml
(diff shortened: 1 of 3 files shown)


conf/.gitignore
```diff
@@ -0,0 +1 @@
+/config.xml
```
tomas-pluskal: I don't think you should add this file - if you want, you can add `/conf/config.xml` to the `.gitignore` file in the root folder

tomas-pluskal:
Thanks, Steffen!

Issue #696, opened 2019-09-04
feature finding error using Waters DDA data


katherinemaloney:
### Basic information

* My operating system: Windows
* My MZmine version: 2.41.2

### What happened

* I'm trying to do use MZmine2 to generate a feature quantification table for use to do [feature-based molecular networking](https://ccms-ucsd.github.io/GNPSDocumentation/featurebasedmolecularnetworking-with-mzmine2/). I'm able to do the mass detection step, but when I try to use the ADAP chromatogram builder, I get the following error (for each of my mzML files): "Retention time of scan #47 is smaller then the retention time of the previous scan. Please make sure you only use scans with increasing retention times. You can restrict the scan numbers in the parameters, or you can use the Crop filter module"

As I look through the scans in each file, I notice a similar pattern:
The first scan is always an MS2 scan.
Scans 2 through ~45 are MS1 scans spaced ~every 0.1 minute apart.
Scans ~46 through ~260 are MS1 scans spaced ~every 0.03 minutes apart
The remaining scans are MS2 scans, roughly - but not exactly - in retention time order.

![image](https://user-images.githubusercontent.com/54910364/64281931-1ece4280-cf09-11e9-80c4-9fefd8ecee67.png)

I was going to try and use the crop filter to select a set of in-order scans, but I'm not sure (a) which set of scans I should be using for the chromatogram builder - presumably the second set since they are spaced closer together? or (b) how to set up the crop filter given that the exact scan number where the retention times restart varies slightly from sample to sample (MS1 restarts at scan# 45-47, switches to MS2 at scan# 250-262).

Or is it a problem with the raw file conversion to mzML?

I was hoping to upload the project file here, but now it seems that the program has frozen and I can't save it. :/

asmirn1:
Hi,

Could you upload and share an mzML data file? 

First, we probably should understand why your data has two sets of MS1 scans.

katherinemaloney:
Hmm. I'd like to but I can't quite figure out how... Tried creating a repository, but it said the file was too big. Is there another way to share it?

asmirn1:
You can upload file on Google Drive or Dropbox and post a link here

katherinemaloney:
Thanks for the speedy reply. Here goes: https://drive.google.com/file/d/14zec5OsCVI7ba73ErpKsFlWux9oZ-M61/view?usp=sharing

asmirn1:
I've looked into your file, and the order of scans in the file is different from their order in MZmine.

[...]

PR #699, opened 2019-09-06
Fix issue #696

asmirn1:
Tomas,

Take a look at this solution. It correctly reads the file from #696 now.


tomas-pluskal:
Hi Alex,
I think this implementation has one drawback that the `unmarshaller.unmarshalCollectionFromXpath()` function is called twice, so essentially the file will be traversed twice. But perhaps it is okay for now, because soon we are going to swap the mzML parser for the one that is included in MSDK, anyway.

PR #702, opened 2019-09-10
Fixed crash after sum formula prediction, inability to reopen parameter setup windows in DPP

SteffenHeu:
Hi Tomas,

this PR fixes a null pointer exception that occured, when using the sum formula prediction in data point processing. This problem occurred when the DataPointsDataSet checks for zero/double data points and recreates them from a list as DataPoint[0]. I don't know exactly why this happened, but it should be fixed now.
Furthermore, the results data set of the sum formula prediction will only contain data points that did get an actual sum formula assigned to them.
Also, i fixed a small bug that occured when a parameter setup window in the data point processing component was closed without pressing O.K. and resulted in it not being able to be reopened again until you closed the whole component.

Best regards
Steffen


tomas-pluskal:
Thanks Steffen!

PR #706, opened 2019-09-18
Encrypted parameters

mfleisch:
### Implementation of parameter encryption and sensitive parameter handling
* Parameters can be saved AES encrypted using a user/config specific encryption key
* Parameters can be sensitive. Sensitive Parameters will not be written into project configuration.xml (e.g. Passwords)
* Changed the location of the mzmine config.xml file to the user home directory

#### Sensitive Parameters:
* Error email config is declared as sensitive
* GNPS login credential are declared as sensitive
* The PasswordParameter is encrypted and sensitive

### DIFF
--- a/src/main/java/net/sf/mzmine/parameters/impl/SimpleParameterSet.java
+++ b/src/main/java/net/sf/mzmine/parameters/impl/SimpleParameterSet.java
@@ -18,21 +18,20 @@
 
 package net.sf.mzmine.parameters.impl;
 
-import java.awt.Window;
-import java.util.Arrays;
-import java.util.Collection;
-import java.util.logging.Level;
-import java.util.logging.Logger;
-
-import org.w3c.dom.Document;
-import org.w3c.dom.Element;
-import org.w3c.dom.NodeList;
-
 import net.sf.mzmine.main.MZmineCore;
 import net.sf.mzmine.parameters.Parameter;
 import net.sf.mzmine.parameters.ParameterSet;
 import net.sf.mzmine.parameters.dialogs.ParameterSetupDialog;
 import net.sf.mzmine.util.ExitCode;
+import org.w3c.dom.Document;
+import org.w3c.dom.Element;
+import org.w3c.dom.NodeList;
+
+import java.awt.*;
+import java.util.Arrays;
+import java.util.Collection;
+import java.util.logging.Level;
+import java.util.logging.Logger;
 
 /**
  * Simple storage for the parameters. A typical MZmine module will inherit this class and define the
@@ -46,6 +45,7 @@ public class SimpleParameterSet implements ParameterSet {
   private static final String nameAttribute = "name";
 
   private Parameter<?> parameters[];
+  private boolean skipSensitiveParameters = false;
 
   public SimpleParameterSet() {
     this.parameters = new Parameter<?>[0];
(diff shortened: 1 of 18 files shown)


src/main/java/net/sf/mzmine/util/StringCrypter.java
```diff
@@ -0,0 +1,123 @@
+package net.sf.mzmine.util;
```
tomas-pluskal: Could you please copy the license header to all new .java files?

src/main/java/net/sf/mzmine/parameters/parametertypes/PasswordParameter.java
```diff
@@ -83,7 +87,7 @@ public void setValue(String newValue) {
   public boolean checkValue(Collection<String> errorMessages) {
     if (!valueRequired)
       return true;
-    if ((value == null) || (value.trim().length() == 0)) {
+    if ((value == null) || (value.isEmpty())) { //white spaces can be valid in pw. So no trim
```
tomas-pluskal: It's true that pw can contain white space, but don't you think a password that consists only of whitespace should be considered invalid? :)

src/main/java/net/sf/mzmine/util/StringCrypter.java
```diff
@@ -0,0 +1,123 @@
...
+
+
+import de.unijena.bioinf.babelms.utils.Base64;
```
tomas-pluskal: Could you switch this to use java.util.Base64 class instead the de.unijena class?

src/main/java/net/sf/mzmine/parameters/impl/SimpleParameterSet.java
```diff
@@ -59,6 +60,14 @@ public SimpleParameterSet(Parameter<?> parameters[]) {
     return parameters;
   }
 
+  public void setSkipSensitiveParameters(boolean skipSensitiveParameters) {
```
tomas-pluskal: Is it useful to have the skipSensitiveParameters as a field of this class (i.e., a state that is remembered)? I would simply add it as a parameter of saveValuesToXML()

tomas-pluskal:
Thanks Markus, it looks good. I added a few minor comments.

src/main/java/net/sf/mzmine/parameters/impl/SimpleParameterSet.java
mfleisch: I agree that Parameter and ParameterSet objects should be kept lightweight.

But if we just add it as a parameter of saveValuesToXML(), it is not so easy to inherit this information to Parameters that contain ParameterSets (e.g. ParameterSetParameter).

To do so we would have to add the skipSensitive boolean to the saveValueToXML() method of the Parameters.

The disadvantage of this solution is that we would have to check the skipSensitive value twice. One time in the ParameterSet (to not write an empty xml block of skipped parameters) and another time within the saveValueToXML() method of sensitive parameters. Which I found less consistent.

In general, I am fine with both solutions. Let me know which you prefer.

src/main/java/net/sf/mzmine/util/StringCrypter.java
mfleisch: Done :-)

src/main/java/net/sf/mzmine/parameters/parametertypes/PasswordParameter.java
mfleisch: I thought that it is a bit strange to check the password requirements of third party applications.
However, all whitespace passwords are a strange edge case. I reverted the change.

src/main/java/net/sf/mzmine/util/StringCrypter.java
mfleisch: Of course, this was by accident :-)

tomas-pluskal:
Looks good, thanks!

PR #707, opened 2019-09-19
A small change for an MSP Export parameter

asmirn1:
This is a small change of the parameter that determines whether the MSP Export outputs integer or fractional m/z values. 

I've replaced two parameters (BooleanParameter, ComboParameter<String>) with one optional parameter. Also, I've change the default value.


tomas-pluskal:
Looks good, thanks!

Issue #711, opened 2019-09-27
Visualization -- Van Krevelen Diagram


Mingruis:
### Basic information

* My operating system: 64-bit operating system, x64-based processor
* My MZmine version: MZmine 2.51

### What happened: After I have done all my preprocessing (Mass detection, FTMS shoulder peaks filter, Chromatogram builder, Chromatogram smoothing,  Chromatogram deconvolution, Isotopic peaks grouper, RANSAC aligner, m/z and RT gap filter, Duplicate peak filter and formula prediction), I'd like to draw Van Krevelen Diagram.
 
For the Diagram, I used these parameters: Peaks: all; Z-Axis: m/z; Z-Axis scale: percentile; Range: 0.5-99.5; Heatmap style: Rainbow. However, there always an error message: " Nothing to plot here or some peaks have other identities than sum formulas. Have you annotated your features with sum formulas? You can use the feature list method-Formula prediction to handle the task." Here attached the screenshot: 
![image](https://user-images.githubusercontent.com/54156546/65733735-d37cff00-e084-11e9-8006-fb338a1b350f.png)

After this error message, I checked the Formula prediction step and deleted all the peaks without formula. And then, redid the Van Krevelen Diagram and the error massage appeared again.

 

* Please describe what happened and how it differs from what you expected to happen.
* Please include all parameter values for the MZmine modules you used. 
* Please upload screenshots of any errors and your log/mzmine.log file.
* Please upload an MZmine project file that demonstrates the problem you are reporting
![error](https://user-images.githubusercontent.com/54156546/65733952-b137b100-e085-11e9-9222-e8a09dc066ec.png)
.

Annexhc:
Hi Mingruis,
I have just checked the module for the current version with another dataset and it seems to work fine. Anyway, can you upload your mzmine project so I can test it (Dropbox, Google drive etc.)? I am traveling today, I will check it tomorrow. You can send me the link via email (ansgar.korf@wwu.de).

Annexhc:
can be closed

PR #718, opened 2019-10-13
Van Krevelen fix

Annexhc:
Hi Tomas,
this PR addresses issue  #711. I have improved the handling of suitable and not suitable molecular formulas for Van Krevelen diagrams. If a user now tries to plot molecular formulas without the elements C,H, and O, as it happend in issue #711, he will get an information message and some warnings in the log file. 
best wishes,
Ansgar


tomas-pluskal:
Looks good!

PR #719, opened 2019-10-16
Join aligner spectra similarity

Annexhc:
Dear Tomas,
I have added spectra similarity as an optional parameter to the join aligner following @robinschmid suggestion to potentially solve the problems described in issue #627.  
The user can now compare the similarity of the representative MS1 scans or the similarity of best MS2 scans. I have tested it with LC-MS/MS data and GC-EI-TOF MS data. General alignment results seem pretty good.
best Ansgar

### DIFF
--- a/src/main/java/net/sf/mzmine/modules/peaklistmethods/alignment/join/JoinAlignerTask.java
+++ b/src/main/java/net/sf/mzmine/modules/peaklistmethods/alignment/join/JoinAlignerTask.java
@@ -249,7 +249,12 @@ class JoinAlignerTask extends AbstractTask {
 
             // compare mass list data points of selected scans
             if (rowDPs != null && candidateDPs != null) {
+
+              // calculate similarity using SimilarityFunction
               sim = createSimilarity(rowDPs, candidateDPs);
+
+              // check if similarity is null. Similarity is not null if similarity score is >= the
+              // user set threshold
               if (sim == null) {
                 continue;
               }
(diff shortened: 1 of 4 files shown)


src/main/java/net/sf/mzmine/modules/peaklistmethods/alignment/join/JoinAlignerTask.java
```diff
@@ -196,16 +223,46 @@ public void run() {
...
+                rowDPs = row.getBestFragmentation().getMassList(massList).getDataPoints();
+                candidateDPs =
+                    candidate.getBestFragmentation().getMassList(massList).getDataPoints();
+              } else
+                continue;
+            }
+
+            // compare mass list data points of selected scans
+            if (rowDPs != null && candidateDPs != null) {
+              sim = createSimilarity(rowDPs, candidateDPs);
```
tomas-pluskal: So, I see that _sim_ is assigned a value here, but I don't see where this value is used? In other words, where in the code do you actually take the calculated spectral similarity into account?

tomas-pluskal:
See my comment to the code. Also, shouldn't there be a parameter for a threshold (what similarity level is considered okay and what is not okay anymore)?

Annexhc:
Okay, the code is not really understandable without seeing the package net.sf.mzmine.util.scans.similarity. 
I have added an optional module parameter to the join aligner parameter set. If checked, it allows to set up JoinAlignerSpectraSimilarityScoreParameters. That parameter set has parameter ModuleComboParameter SpectralSimilarityFunction. There the user can choose between WeightedCosineSpectralSimilarity and CompositeCosineSpectralSimilarity. A third parameter set, including the threshold for the similarity, will be created. 
To answer the comment above, the sim value is used. A similarity will only be assigned if all requirements were met(net.sf.mzmine.util.scans.similarity.SpectraSimilarityFunction), otherwise it will be null. Thats why I only checked for null.
join_aligner_sim

tomas-pluskal:
Okay! Perhaps you could add this explanation to the code?

tomas-pluskal:
Thanks!

Issue #729, opened 2019-10-22
Spectral export check button


tomas-pluskal:
It is unclear what this button actually does. Perhaps some tooltip would be useful?

<img width="1487" alt="Screen Shot 2019-10-22 at 11 23 02" src="https://user-images.githubusercontent.com/1796692/67302154-7414e200-f4be-11e9-9f05-50ad2473acf6.png">

robinschmid:
Solved by PR #732 (?)

Button is removed when only one entry is submitted. 
Tooltips added to all buttons and text fields

PR #732, opened 2019-10-23
Spectral library submit: Tooltips, disable buttons ... 

robinschmid:
Hey Tomas,

this PR refers to your issue #729.

- added tooltips to all buttons and text fields
- exclude button (check button) is now removed if the user only exports one entry at a time
    - this button allows to exclude single entries, if multiple ions of the same compound are exported at once
- next and prev buttons now have disabled icons and are disabled accordingly

image


tomas-pluskal:
Thanks Robin!

PR #734, opened 2019-10-24
Agilent ICP raw data import and  manual Integration via XIC

SteffenHeu:
Hi Tomas, 
this PR contains an implementation of an Agilent ICP raw data file import and allows the user manual integration via an XIC plot rather than just a popup window with m/z and rt parameters. I could not get my hands on a file that would be loaded using the current Agilent csv import module, but I think the discrimination in the RawDataFileTypeDetector should work.

The integration window looks like this:
manualintegration
The user can either click in the XIC to set the boundaries or use the text boxes on the bottom. The window will draw the current boundaries in green and red and also draw a line on the mouse pointer in the colour, of the currently selected boundary. 
However, i removed the m/z range parameter for this window, since it will only be opened from an already created peak list, so the m/z range should be set already. 
It can be opened by either doing the regular right click -> manually define peak or by double clicking on the area column in a peak list. This also works for aligned peak lists.

If you have any further suggestions, I'm always up to implement them.
Best regards
Steffen


SteffenHeu:
Also: Please note the out-commented line in the peak row summary window. Was there a reason for this? I couldn't find any module that does not work with multiple selections in there, yet it was being forbidden. I think removing that makes it a lot more comforttable to export the data to excel or something, since you can just copy+paste the whole table.

src/main/java/net/sf/mzmine/modules/rawdatamethods/rawdataimport/fileformats/CsvReadTask.java
```diff
@@ -0,0 +1,202 @@
...
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
+import java.util.Scanner;
+import java.util.logging.Level;
+import java.util.logging.Logger;
+import javax.annotation.Nonnull;
+import javax.annotation.Nullable;
+import com.google.common.collect.Range;
+import de.unijena.bioinf.ChemistryBase.ms.SpectrumProperty;
```
tomas-pluskal: I wonder why this class needs to import the de.unijena package?

src/main/java/net/sf/mzmine/modules/rawdatamethods/peakpicking/manual/XICManualPickerModule.java
```diff
@@ -0,0 +1,109 @@
...
+import net.sf.mzmine.main.MZmineCore;
+import net.sf.mzmine.modules.MZmineModule;
+import net.sf.mzmine.modules.visualization.peaklisttable.table.PeakListTable;
+import net.sf.mzmine.parameters.ParameterSet;
+import net.sf.mzmine.parameters.parametertypes.selectors.RawDataFilesSelection;
+import net.sf.mzmine.parameters.parametertypes.selectors.RawDataFilesSelectionType;
+import net.sf.mzmine.util.ExitCode;
+import com.google.common.collect.Range;
+
+public class XICManualPickerModule implements MZmineModule {
```
tomas-pluskal: This new modules should be listed in the MZmineModulesList class.

tomas-pluskal:
Thanks Steffen, it looks good, I just added a few minor comments.

For next time, please try to submit each change as a separate pull request. Since this one combines changes to the raw data import, the manual peak picker, and the peak summary component, it is much harder to review.

tomas-pluskal:
👍🏻

PR #738, opened 2019-10-28
Fix NullPointer ADAP hierarchical clustering

robinschmid:
Fix for row.getPeakInformation() == null after FeatureListRowFilter

Fix for nullpointer in ADAP hierarchical clustering


tomas-pluskal:
I believe row.getPeakInformation() was added for the modules developed by @du-lab, but it is not well documented. I agree with you that we should change it to something more meaningful.

Also putting copyPeakRow() into utils sounds good.

PR #739, opened 2019-10-28
Option to remove empty rows after PeakFilter

robinschmid:
Hey Tomas,

this is a small addition to the PeakFilter module. If the resulting feature list row is empty - it can now be removed automatically. Previously, we had to run PeakFilter + FeatureListRowFIlter (min features=1) to do this.

Cheers
Robin


tomas-pluskal:
Hmm, maybe removing empty rows should just be enforced in all cases? Seems like it serves no purpose to keep empty rows.

robinschmid:
Yeah this was my first plan but then I decided for the "compatibility" version.

Gonna change it to always - its a better user experience.

robinschmid:
Done.

tomas-pluskal:
Thanks Robin!

PR #749, opened 2019-11-08
[GNPS][ADAP] Workflow for GC-MS molecular networking with Kovats Index

robinschmid:
Hey all,

This is a preview. Not ready for merging.

I am working on an export of ADAP workflow results to the GC-MS workflow on GNPS.

### Done
- Export of quant table
- Export of an mgf spectra file (ADAP clustered spectra)
- Added a dialog for Kovats retention time picking
    - See first video: https://www.youtube.com/watch?v=XodHMJcuwnk&t=2s
    - Combined base-peak plot of up to 2 raw data files
        - Direct selection of m/z for specific alkane signals
    - Peak picking and annotation as alkane standards
    - Save to/load from files
    - Mouse gesture to correct retention time markers

### To-Do
- Finalize formats on GNPS side
- Direct submission to GNPS GC-MS (similar to FBMN connection of MZmine->GNPS)
- See issue : Adding Kovats Index to data model and to local spectra library matching

image


tomas-pluskal:
Hi Robin,

This looks like a great addition! Are you coordinating the effort with @du-lab and @asmirn1? They are probably the most experienced in using MZmine for GC-MS and could give useful suggestions.

My concerns with the Kovats retention time picking module: it relies heavily on user interaction (in other words, it cannot be executed as part of batch processing). It also duplicates the function of other modules for peak detection. My preferred way would be to use existing modules for feature detection and (for example) add a new module for identification of Kovats index-specific peaks, which would use the existing mechanisms for tracking peak identity and for exporting the results. It is probably too late to change the design now, but let's keep this in mind for the future.

robinschmid:
It was a quick addition so far. Up to now, contact was only about the ADAP-mgf export changes.
However, input or opinions on the new tool are more than appreciated. @du-lab @asmirn1 

This tool is intended to have people check the results and give manual adjustment options. Once the Kovats retention time table is exported, it can be included in batch processing. My plan was to add an automatic module based on the feature lists later on. However, there are some problems to address to fully automate this. Manually adjusted results will be better. Considering that people do not have many Kovats standard runs, the workflow is only a marginal portion of the whole processing workflow.

asmirn1:
Hi Robin,

It's a good addition to MZmine. As I understand, this module helps find alkanes in a Kovats mixture sample. Is it possible to use these results to calculate the retention index within MZmine? It would be nice to have a module for calculating retention indices.

robinschmid:
We still need to add the retention index to the data model and a module to calculate it from the alkane/retention time list - which we export to a csv file.
Do you already have the java algorithms to calculate it?

asmirn1:
No, I don't have a java implementation, but it should be fairly simple (the formula is [here](https://en.wikipedia.org/wiki/Kovats_retention_index)). As you said, the main work is to add new data structures and a module. I guess, we can import alkanes from the csv file that you create.

tsufz:
This is a great enhancement, I am not so familar with newer GC-MS software. I know it only from `AMDIS`. It would be helpful, if the module is quite open for the training of different models. In the LC-MS field are different developments ongoing. In `MassBank`, we already implemented two RTI for LC (see https://github.com/tsufz/MassBank-web/blob/dev/Documentation/MassBankRecordFormat.md).

asmirn1:
@tsufz If you are interested in GC-MS algorithms, [here is a tutorial](http://mzmine.github.io/ADAP_user_manual.pdf) that goes through main steps of processing GC-MS in MZmine.

minzaroma:
@robinschmid Having requested this feature in the past, let me just drop in to say I just tried it and it looks very promising to me!
I ran a C10-40 alkane ladder, and all peaks were easily identified by the module without any issues, despite some small "dirt peaks" appearing in between. ;) If I have it correctly, this picture now shows the overlay of my alkane standard and the other sequence selected: 
grafik
I appreciate the robust performance and the visualization of the module - seems you do it right! ;) As I understood, this is now more of a "preview feature", or can I already use the results in some way?

robinschmid:
That's a nice feedback. So far you can only use it for the GC-MS workflow on GNPS. 
We are currently implementing the Kovats index to the library matching workflow in MZMine

minzaroma:
Very cool, great news! Can't wait to use it in the future :D

PR #754, opened 2019-11-14
Data point processing fix - Null pointer exception on first start

SteffenHeu:
Hi tomas,
i noticed a null pointer exception on the first start of the data point processing task, when some if the parameters have not been set, which caused a null value in a boolean parameter somehow. The parameter sets are now checked before the creation of the task, which fixes the problem.

Best
Steffen


tomas-pluskal:
Sounds good, thanks!

PR #755, opened 2019-11-14
Any element deisotoping for data point processing

SteffenHeu:
Hi Tomas,

before the announced change I wanted to commit this module. It basically works, but not totally as I would want it to work. At the moment it searches for isotopic mass differences for more than just 13C peaks and generates an isotope pattern from that. The result can than be used automatically as a basis for a sum formula prediction during an DataPointProcessingQueue.

I think this would be a good addition in the future, but at the moment there are some things, the module cannot handly yet:
- does not work if not all isotopic peaks are present
- takes the monoisotopic/lowest mass as representative peak, not the most abundant. Furthermore, for the sum formula prediction this might mean, that the formula might be calculated for the wrong peak, since it uses the major isotope formula. Maybe that is something to think about in the future.

This module is currently not listed in the module list, so it cannot be used in the current state.

Best regards
Steffen


tomas-pluskal:
Thanks Steffen! It sounds like a useful addition, please continue working on it after the codebase refactoring.

Issue #742, opened 2019-10-29
How to remove features that exist in blanks as well after alignment? 


QizhiSu:
### Basic information
Hi everybody, I am thinking if it's possible to remove features that exist in blanks as well by setting a fold change parameter? For example, those features that have peak height/area 10 times than higher than that in the blank will be kept, and the rest will be removed.

Many thanks in advance.

* My operating system:
* My MZmine version: 

### What happened

* Please describe what happened and how it differs from what you expected to happen.
* Please include all parameter values for the MZmine modules you used. 
* Please upload screenshots of any errors and your log/mzmine.log file.
* Please upload an MZmine project file that demonstrates the problem you are reporting.

SteffenHeu:
Hi Sukis,
I am not aware of a module like that at the moment. However, I am currently working on a module that would be able to "substract" a peak list from another. The fold change seems like a good idea to do as well. But the implementation is by no means finished yet, but I can let you know here.

Sukis123:
@SteffenHeu Wow great! Many thanks.

robinschmid:
Yeah, I just wanted to link out to @SteffenHeu but he was way faster than me ;)

This feature was already requested a couple of times (@lfnothias, @Annexhc ).
Maybe we can use this thread to come up with a good strategy:

### Blank removal
- use blank sample(s) to cut out **insignificant** features
- **insignificant**: Below fold-change (sample/blank < threshold) 
- Easy: Use blank features
- Comprehensive: Use blank chromatograms (to tackle issues 1 and 2)
    - highest intensity in retention time window

### Possible issues:
#### 1. Duplicates in the aligned list: 
- A feature in the blank sample can only be aligned to one feature list row. 
-  we need to make sure that the blank feature is subtracted from all "duplicates" 

#### 2. Features from noisy background signals 
- usually spanning the whole chromatograms
- random retention time
- unlikely to match blank sample even if signal intensities are on the same levels

Annexhc:
Hey guys,
[...]

PR #756, opened 2019-11-18
Feature list subtraction - unlisted

SteffenHeu:
This is another module I am currently working on, but i did not have time to implement all the suggestions mentioned in #742, which is why the module is currently not listed in the modules list.


tomas-pluskal:
Sounds good, thanks!

Issue #762, opened 2019-11-22
"Error of task ADAP Peak decompositon"


minzaroma:
### Basic information

* My operating system: Windows 2012 Server R2 Standard
* My MZmine version: 2.52

### What happened

Hey all,

I was analyzing some GC-MS data using the new MZmine version (2.52). Unfortunately, the ADAP spectral deconvolution (hierarchical clustering) module failed with the following note: "cern/colt/matrix/DoubleMatrix1D".  Here's a screenshot:

![grafik](https://user-images.githubusercontent.com/52671757/69435268-84c7aa80-0d3f-11ea-94f3-783b04659dee.png)

Seems to be a new new issue as I never had it in earlier versions (2.4....).

Cheers, Caro

tomas-pluskal:
Looks like a missing dependency or something... @asmirn1 ?

tomas-pluskal:
@minzaroma Could you provide your project file so we can reproduce the issue?

minzaroma:
Yes, but it's a bit big

tomas-pluskal:
Use Dropbox, Google Drive, or something similar.

minzaroma:
Ok, here it is: https://drive.google.com/open?id=1b7iONTTVTC9lIMk4VTeuBIR-GAwvYne2
Thanks for your fast response!

asmirn1:
It does look like a dependency issue. 

Thank you for the project file. I'll try to reproduce the error.

asmirn1:
Hi @minzaroma I've made some changes to the developer version of MZmine. Here is the updated version: https://drive.google.com/file/d/1dhQRkGCpbAT2xYC7yGoR4_r7FksqsXUF/view?usp=sharing

Could you try it and let me know if it works. If it does, we'll push these changes to the next release version of MZmine.

minzaroma:
Thanks @asmirn1 I tried your version just now and it seems to be all fine again!

Cheers, Caro

tomas-pluskal:
Should be fixed in version 2.53

PR #766, opened 2019-12-02
Update the version of ADAP. The new jar file includes all dependencies.

asmirn1:
Fix for issue #762


tomas-pluskal:
Hi Alex,
I am not super happy with this solution. Your ADAP jar has now become an "uber-jar" that contains many libraries that are already included with MZmine as standard dependencies (SLF4J, Apache Commons, Guava, etc.). This can create problems because classes on the classpath may be duplicated with different versions that may not work well together and create rare and hard-to-track exceptions.

The dependencies should instead be declared in your pom.xml so Gradle can create a dependency graph.

tomas-pluskal:
Also, we moved the main development to the https://github.com/mzmine/mzmine3/ repository. I may still make one or two bug-fix releases of MZmine 2 from this repository, but the main effort is now focusing on MZmine 3.

robinschmid:
Is the ADAP library on github?
Maybe it is possible to use https://jitpack.io/ to include the ADAP libraries into MZmine. What do you think?

asmirn1:
Hi Tomas, it seems that after switching from Maven to Gradle, gradle couldn't pick up the dependencies that ADAP requires and that would cause the issue #762. That's why I wanted to pack everything into our jar file.

I can pack only specific packages into our jar file instead.

asmirn1:
@robinschmid ADAP is currently on Bitbucket. I'm not familiar with JitPack, let me take a look.

tomas-pluskal:
I think gradle should pick up all the transitive dependencies, but maybe it doesn't pick up the repository definitions. So if you are loading dependencies from some other repositories, you may need to list them in the build.gradle.

asmirn1:
Okay, I'll try the master version again and see why Gradle doesn't pick up ADAP dependencies.

Also, ADAP library is still on Java 8, so it doesn't use java modules, etc. Maybe, that's the reason.

tomas-pluskal:
MZmine also doesn't use java modules, so that is fine.

tomas-pluskal:
Try running `./gradlew -q dependencies`

tomas-pluskal:
I found what the problem is - it's this declaration in the build.gradle:
````
        metadataSources {
            artifact()
        }
````
I added it a while ago because the builds were failing, but now it seems to build fine without this declaration. So I will fix it.

**Comment by tomas-pluskal** [comment]
How badly is ADAP broken in MZmine 2.52? Should I make a new release with the correct dependencies?

**Comment by asmirn1** [comment]
So far, the only module that would give an error is Spectral Deconvolution/Hierarchical Clustering. I'll check other modules.

**Comment by asmirn1** [comment]
Tomas,

Spectral Deconvolution/Multivariate Curve Resolution doesn't work either. It cannot find jsparsehc, which is in the maven central repository: https://mvnrepository.com/artifact/org.du-lab.jsparsehc/jsparsehc. 

FYI, when I try to compile mzmine without 
```
metadataSources {
            artifact()
        }
```
I get an error: 
[error.txt](https://github.com/mzmine/mzmine2/files/3913263/error.txt)

tomas-pluskal:
Please pull the latest master, I added the missing .pom files. It should build fine now. Perhaps we should release the version 2.53 including the fix then?

asmirn1:
@tomas-pluskal I've tested the latest master branch and it works great, no errors for any ADAP modules. Yes, making a new release would be great!

PR #784, opened 2020-01-20
Fix for legend not matching graph colour after applying theme during export

SteffenHeu:
Hi Tomas, 
this fixes the issue as described above and discussed in Slack.
Also, the export uses the colour preset as set in the preferences. I thought that might be useful.

Best regards
Steffen


tomas-pluskal:
Hi @SteffenHeu,
Thanks for the fix! I’ll merge it but can you also please submit it to the mzmine3 repo?

Issue #782, opened 2020-01-17
New manual graphical peak integration


Gregor90:
### Basic information

* My operating system: Win10
* My MZmine version: 2.53

### What happened
Hi Tomas,

I'm a great fan of the newly added manual graphical peak integration (with MZmine2 V2.52), but unfortunately in this new function it is not possible to manually adjust the m/z value. Please re-introduce this feature with the next update. 

Best

Gregor

PR #786, opened 2020-01-23
Manual Integration m/z selection

SteffenHeu:
Hey Tomas,
As requested in #782, i readded the manual selection of the m/z range to the manual integration window.

Best regards
Steffen


tomas-pluskal:
Thanks @SteffenHeu !
Would you like to volunteer to also update this module in the mzmine3 / JavaFX branch? :)

SteffenHeu:
I can do that, yes, but i think for MZmine 3 I would recode the dialog, so the m/z range can be set graphically as well.
I just remembered that a graphical and a keyboard input into the text field was a pain to implement in the first place, so i would somehow go for a different solution in MZmine 3.

SteffenHeu:
And I think it might be smart to implement the colour palettes into the preferences and the various plots before that.

PR #794, opened 2020-02-09
.gitignore update for idea ide

SteffenHeu:
Hi,

I updated the .gitignore so it does not update the idea ide files to github.

Best regards
Steffen


tomas-pluskal:
Thanks! But I think we should all focus our efforts on the mzmine3 repo now :)

Issue #803, opened 2020-03-15
Unhandled exception in ADAPChromatogramBuilderTask.java


bjonnh:
### Basic information

* My operating system: Linux 64
* My MZmine version: 2.53 (release)

### What happened

* Please describe what happened and how it differs from what you expected to happen.

ADAP chromatogram builder fails with message:
19:18:08|SEVERE|io.github.mzmine.taskcontrol.impl.WorkerThread]: Unhandled exception java.lang.ArrayIndexOutOfBoundsException: Index 0 out of bounds for length 0 while processing task Detecting chromatograms in 021_1-16_1_61.mzML filtered filtered(java.lang.ArrayIndexOutOfBoundsException: Index 0 out of bounds for length 0@io.github.mzmine.modules.dataprocessing.featdet_ADAPchromatogrambuilde
r.ADAPChromatogramBuilderTask.run(ADAPChromatogramBuilderTask.java:258))                                                                                                                             
java.lang.ArrayIndexOutOfBoundsException: Index 0 out of bounds for length 0                                                                                                                         
        at io.github.mzmine.modules.dataprocessing.featdet_ADAPchromatogrambuilder.ADAPChromatogramBuilderTask.run(ADAPChromatogramBuilderTask.java:258)                                             
        at io.github.mzmine.taskcontrol.impl.WorkerThread.run(WorkerThread.java:60)                                                                                                                  
[19:18:08|SEVERE|io.github.mzmine.gui.impl.HeadLessDesktop]: Unhandled exception in task Detecting chromatograms in 021_1-16_1_61.mzML filtered filtered: java.lang.ArrayIndexOutOfBoundsException: Index 0 out of bounds for length 0 (ADAPChromatogramBuilderTask.java:258)             

* Please include all parameter values for the MZmine modules you used. 

See:
https://www.bjonnh.net/share/batch_queue_parameters_v01.xml

* Please upload screenshots of any errors and your log/mzmine.log file.
```
openjdk version "12.0.2" 2019-07-16
OpenJDK Runtime Environment (build 12.0.2+10)
OpenJDK 64-Bit Server VM (build 12.0.2+10, mixed mode)
[17:09:53|INFO|io.github.mzmine.main.MZmineCore]: Starting MZmine 2.53
[17:09:53|FINE|io.github.mzmine.main.TmpFileCleanup]: Checking for old temporary files...
[17:09:53|FINE|io.github.mzmine.main.MZmineCore]: Loading core classes..
[...]

PR #804, opened 2020-03-15
Solve Issue #803 Add a check if ADAP didn't find any peak

bjonnh:
Avoid the error explained in issue #803 .

Returns with a status Finished.


tomas-pluskal:
Looks good, thanks! 
Could you please also submit the same fix to the mzmine3 repository (https://github.com/mzmine/mzmine3)? I don't expect any more releases to come from the mzmine2 repo, actually.

PR #859, opened 2020-09-18
add support for NetCDF-4/HDF5 version of CDF

### LINKED ISSUE #850: cdf with HDF-format
### Basic information

* My operating system: kubuntu 20.04
* My MZmine version: 2.53

### What happened

Hello,
there is file type detection in the source code https://github.com/mzmine/mzmine2/blob/master/src/main/java/net/sf/mzmine/modules/rawdatamethods/rawdataimport/RawDataFileTypeDetector.java#L37.

I have a netCDF-file with "HDF" in its header.

If I understood correctly in case of using netCDF library from unidata.ucar, it is possible to work with both CDF and HDF formats of netCDF.

Is there any possibility to support HDF format in MZmine?

zdens:
closes #850 

These changes were tested via building from source on my Kubuntu 20.04. 

For testing there were used Thermo Finnigan RAW-file, CDF-file obtained from Xcalibur software and another CDF-file obtained using Scientific Data Set library shipped with NetCDF v.4.0.1.0.

### DIFF
--- a/src/main/java/net/sf/mzmine/modules/rawdatamethods/rawdataimport/RawDataFileTypeDetector.java
+++ b/src/main/java/net/sf/mzmine/modules/rawdatamethods/rawdataimport/RawDataFileTypeDetector.java
@@ -21,6 +21,7 @@ package net.sf.mzmine.modules.rawdatamethods.rawdataimport;
 import java.io.File;
 import java.io.FileInputStream;
 import java.io.InputStreamReader;
+import java.nio.charset.StandardCharsets;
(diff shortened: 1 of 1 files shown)


tomas-pluskal:
Thanks a lot!

I am going to merge this now, but actually we are not planning to make any new releases from this repository. Instead, we are now working on MZmine 3 in this repo: https://github.com/mzmine/mzmine3

I have ported your code to the MZmine 3 repo as well, so it will be included.

